PR #1, opened 2022-06-08
Simple struct serialize

choinskib:
The PR is implementing the simple XML serializer for the structs with named fields.

### DIFF
--- a/instant-xml-macros/Cargo.toml
+++ b/instant-xml-macros/Cargo.toml
@@ -10,5 +10,6 @@ proc-macro = true
 
 [dependencies]
 heck = "0.4"
+proc-macro2 = "1.0.39"
 quote = "1.0.18"
 syn = { version = "1.0.86", features = ["full"] }
--- a/instant-xml-macros/src/lib.rs
+++ b/instant-xml-macros/src/lib.rs
@@ -6,15 +6,38 @@ use syn::parse_macro_input;
 
 #[proc_macro_derive(ToXml)]
 pub fn to_xml(input: TokenStream) -> TokenStream {
-    let ast = parse_macro_input!(input as syn::ItemStruct);
+    let ast = parse_macro_input!(input as syn::DeriveInput);
     let ident = &ast.ident;
-    let name = ident.to_string();
+    let root_name = ident.to_string();
+    let mut output: proc_macro2::TokenStream =
+        TokenStream::from(quote!("<".to_owned() + #root_name + ">")).into();
+
+    match &ast.data {
+        syn::Data::Struct(ref data) => {
+            match data.fields {
+                syn::Fields::Named(ref fields) => {
+                    fields
+                    .named
+                    .iter()
+                    .for_each(|field| {
+                        let field_name = field.ident.as_ref().unwrap().to_string();
+                        let field_value = field.ident.as_ref().unwrap();
+                        output.extend(quote!(+ "<" + #field_name + ">" + self.#field_value.to_string().as_str() + "</" + #field_name + ">"));
+                    });
+                }
+                syn::Fields::Unnamed(_) => todo!(),
+                syn::Fields::Unit => {}
+            };
+        }
+        _ => todo!(),
+    };
+
+    output.extend(quote!(+ "</" + #root_name + ">"));
+
     TokenStream::from(quote!(
         impl ToXml for #ident {
             fn write_xml<W: ::std::fmt::Write>(&self, write: &mut W) -> Result<(), instant_xml::Error> {
-                write.write_str("<")?;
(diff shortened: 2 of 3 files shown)


instant-xml-macros/src/lib.rs
```diff
@@ -6,15 +6,51 @@ use syn::parse_macro_input;
...
+        }
+        syn::Data::Struct(ref data) => {
+            match data.fields {
+                syn::Fields::Named(ref fields) => {
+                    fields
+                    .named
+                    .iter()
+                    .for_each(|field| {
+                        let field_name = field.ident.as_ref().unwrap().to_string();
+                        let field_value = field.ident.as_ref().unwrap();
```
choinskib, line 28: Is there a way to avoid creating those two variables here?

instant-xml-macros/src/lib.rs
```diff
@@ -6,15 +6,51 @@ use syn::parse_macro_input;
...
+                        output = TokenStream::from(quote!(
```
djc, line 29: You can use `output.extend()` here (from the `Extend` trait), instead of reserializing everything all the time.

instant-xml-macros/src/lib.rs
djc: Nope, I don't think so.

instant-xml-macros/src/lib.rs
```diff
@@ -6,15 +6,51 @@ use syn::parse_macro_input;
...
+                            ::std::format!("{}<{}>{}</{}>", #output, #field_name, self.#field_value, #field_name)
+                        )).into();
+                    });
+                }
+                syn::Fields::Unnamed(_) => {
```
djc, line 34: You can write a simple branch like this as `todo!(),` instead of `{ todo!(); }`.

djc:
(I've put some basic CI in place in #2 similar to what we have in other repos, so you'll probably have to clean some things up for this to pass CI.)

instant-xml-macros/src/lib.rs
choinskib: Ok, thanks , I changed it

instant-xml-macros/src/lib.rs

instant-xml-macros/src/lib.rs
```diff
@@ -6,15 +6,43 @@ use syn::parse_macro_input;
...
-    let name = ident.to_string();
+    let root_name = ident.to_string();
+    let mut output: proc_macro2::TokenStream =
+        TokenStream::from(quote!(::std::format!("<{}>", #root_name)
+            .as_str()
+            .to_owned()))
+        .into();
+
+    match &ast.data {
+        syn::Data::Enum(_) => todo!(),
```
djc, line 19: Since this is the same as the wildcard case anyway, just leave off this arm?

instant-xml-macros/src/lib.rs
```diff
@@ -6,15 +6,43 @@ use syn::parse_macro_input;
...
+        syn::Data::Struct(ref data) => {
+            match data.fields {
+                syn::Fields::Named(ref fields) => {
+                    fields
+                    .named
+                    .iter()
+                    .for_each(|field| {
+                        let field_name = field.ident.as_ref().unwrap().to_string();
+                        let field_value = field.ident.as_ref().unwrap();
+                        output.extend(quote!(+ ::std::format!("<{}>{}</{}>", #field_name, self.#field_value, #field_name).as_str()));
```
djc, line 29: Doing one `format!()` call per field and concatenating them is very inefficient. I think we should do something better? `ToXml::write_xml()` takes a `fmt::Write`, so we should just write into that directly.

instant-xml-macros/Cargo.toml
```diff
@@ -12,3 +12,4 @@ proc-macro = true
 heck = "0.4"
 quote = "1.0.18"
 syn = { version = "1.0.86", features = ["full"] }
+proc-macro2 = "1.0.39"
```
djc, line 15: Please put this in alphabetic order.

instant-xml-macros/src/lib.rs
choinskib: Good call, done

choinskib:
What gets me thinking about this is that we are creating this big string in the memory, so processing huge XML files might be problematic, it could be maybe rewritten to do it in chunks, with calling the `write.write_str` more often.

choinskib:
What gets me thinking about this is that we are creating this big string in the memory, so processing huge XML files might be problematic, it could be maybe rewritten to do it in chunks, with calling the `write.write_str` more often.

djc (approved):

PR #3, opened 2022-06-09
Default XML namespace support

choinskib:
The PR is implementing simple support for default XML namespace parameter for the structs.

### DIFF
--- a/instant-xml-macros/src/lib.rs
+++ b/instant-xml-macros/src/lib.rs
@@ -5,33 +5,33 @@ use quote::quote;
 use syn::{parse_macro_input, DeriveInput, Lit, Meta, NestedMeta};
 
 fn retrieve_default_namespace(input: &DeriveInput) -> Option<String> {
-    if let NestedMeta::Meta(Meta::List(list)) = get_meta_items(&input.attrs).first()? {
+    for attr in &input.attrs {
+        if !attr.path.is_ident(XML) {
+            continue;
+        }
+
+        let nested = match attr.parse_meta() {
+            Ok(Meta::List(meta)) => meta.nested,
+            Ok(_) => todo!(),
+            _ => todo!(),
+        };
+
+        let list = match nested.first() {
+            Some(NestedMeta::Meta(Meta::List(list))) => list,
+            _ => todo!(),
+        };
+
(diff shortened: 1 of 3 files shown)


instant-xml-macros/src/lib.rs
```diff
@@ -1,19 +1,25 @@
 extern crate proc_macro;
 
+mod common;
```
djc, line 3: I'd prefer not to put this in a separate module for now, since `lib.rs` is still pretty small.

instant-xml-macros/src/common.rs
```diff
@@ -0,0 +1,43 @@
+use syn::{DeriveInput, Lit, Meta, NestedMeta};
+
+pub fn retrieve_default_namespace(input: &DeriveInput) -> Option<String> {
+    for item in get_meta_items(&input.attrs) {
```
djc, line 4: Try to reduce rightward drift by using early `return` or `continue`. Instead of unwrapping `Option`s we can just use `?` on them to return `None` to the caller.

instant-xml-macros/src/common.rs
```diff
@@ -0,0 +1,43 @@
...
+        match item {
+            NestedMeta::Meta(Meta::List(list)) => {
+                if let Some(id) = list.path.get_ident() {
+                    if id == "namespace" {
+                        let nested = list.nested.first().unwrap();
+                        match nested {
+                            NestedMeta::Lit(Lit::Str(v)) => {
+                                let value: syn::Path = v.parse().unwrap();
```
djc, line 12: Why are we parsing a `Path` out of a `Lit::Str` here?

instant-xml-macros/src/lib.rs

instant-xml-macros/src/common.rs
choinskib: I rewritten this function, looks much better now, please take a look.

instant-xml-macros/src/common.rs
choinskib: It was leftover after something, sorry for this PR being a little untidy. I was in rush, but at the same time I wanted to open it yesterday, to gather your thoughts.

djc:
Merged rebased version in #5.

PR #4, opened 2022-06-10
Handling namespaces attributes in fields

choinskib:
This PR is implementing the handle of struct fields namespaces.


instant-xml-macros/src/lib.rs
```diff
@@ -40,8 +70,15 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
...
 
     let mut header: String = root_name.to_string();
-    if let Some(v) = retrieve_default_namespace(&ast) {
-        header += format!(" xmlns=\"{}\"", v).as_str();
+    let (default_namespace, other_namespaces) = match retrieve_namespaces(&ast) {
+        Some((Some(v1), Some(v2))) => (Some(v1), Some(v2)),
+        Some((Some(v1), None)) => (Some(v1), None),
+        Some((None, Some(v2))) => (None, Some(v2)),
+        _ => (None, None),
+    };
```
choinskib, line 78: Couldn't think of any more clever way of retrieving those.

instant-xml-macros/src/lib.rs
```diff
@@ -57,7 +94,24 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
                     .for_each(|field| {
                         let field_name = field.ident.as_ref().unwrap().to_string();
```
choinskib, line 95: This will be extracted to separate fn.

instant-xml-macros/src/lib.rs
```diff
@@ -2,9 +2,39 @@ extern crate proc_macro;
...
+fn retrieve_namespaces(input: &DeriveInput) -> Namespaces {
+    let mut default_namespace = DefaultNamespace::default();
+    let mut other_namesapces = OtherNamespaces::default();
+
+    if let NestedMeta::Meta(Meta::List(list)) = get_meta_items(&input.attrs).first()? {
+        if list.path.get_ident()? == "namespace" {
+            for item in &list.nested {
+                match item {
+                    NestedMeta::Lit(Lit::Str(v)) => {
+                        default_namespace = v.value();
```
choinskib, line 21: Some error handling is needed here for the cases like: 
`#[xml(namespace("URI", "BAZ", foo = "BAR"))]`
As it will make the BAZ default namespace.

instant-xml-macros/src/lib.rs
```diff
@@ -2,9 +2,39 @@ extern crate proc_macro;
...
+                    }
+                    _ => (),
+                }
+            }
+        }
+    }
+    Some((Some(default_namespace), Some(other_namesapces)))
+}
+
+fn retrieve_field_namespace(input: &syn::Field) -> Option<String> {
```
choinskib, line 37: `retrieve_namespaces` and `retrieve_field_namespace` probability could be unified somehow, I will take a look at it after my leave.

instant-xml-macros/src/lib.rs
```diff
@@ -39,9 +84,17 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
...
+
+    let mut header: String = root_name.to_string();
+    let (default_namespace, other_namespaces) = match retrieve_namespaces(&ast) {
+        Some((Some(v1), Some(v2))) => (Some(v1), Some(v2)),
+        Some((Some(v1), None)) => (Some(v1), None),
+        Some((None, Some(v2))) => (None, Some(v2)),
+        _ => (None, None),
+    };
+
+    if default_namespace.is_some() {
```
djc, line 96: Can we do `if let` here?

instant-xml-macros/src/lib.rs
```diff
@@ -22,15 +25,57 @@ fn retrieve_default_namespace(input: &DeriveInput) -> Option<String> {
...
+        if let NestedMeta::Lit(Lit::Str(v)) = list.first()? {
+            return Some(v.value());
+        }
+    }
+    None
+}
+
+type DefaultNamespace = String;
+type OtherNamespaces = HashMap<String, String>;
+type Namespaces = Option<(Option<DefaultNamespace>, Option<OtherNamespaces>)>;
```
djc, line 46: Actually, I'd prefer not using type aliases, as they don't provide extra type safety, and usually just obscure what's actually going on. I also thinks it could make sense to just keep the empty `HashMap` in the case no other namespaces are defined, which should make the code handling the return value simpler.

instant-xml-macros/src/lib.rs
```diff
@@ -22,15 +25,57 @@ fn retrieve_default_namespace(input: &DeriveInput) -> Option<String> {
         };
 
         if list.path.get_ident()? == "namespace" {
-            if let NestedMeta::Lit(Lit::Str(v)) = list.nested.first()? {
-                return Some(v.value());
-            }
+            return Some(list.nested.clone());
```
djc, line 28: Why do we need to `clone()` here?

instant-xml-macros/src/lib.rs
```diff
@@ -2,10 +2,13 @@ extern crate proc_macro;
...
 use proc_macro::TokenStream;
 use quote::quote;
+use std::collections::HashMap;
 use syn::{parse_macro_input, DeriveInput, Lit, Meta, NestedMeta};
 
-fn retrieve_default_namespace(input: &DeriveInput) -> Option<String> {
-    for attr in &input.attrs {
+fn retrieve_namespace_list(
+    attributes: &Vec<syn::Attribute>,
+) -> Option<syn::punctuated::Punctuated<NestedMeta, syn::token::Comma>> {
```
djc, line 10: Maybe this could return an `impl Iterator<Item = NestedMeta>`?

instant-xml/tests/all.rs
```diff
@@ -27,6 +29,6 @@ fn struct_with_named_fields() {
         }
         .to_xml()
         .unwrap(),
-        "<StructWithNamedFields xmlns=\"URI\"><flag>true</flag><string>test</string><number>1</number></StructWithNamedFields>"
+        "<StructWithNamedFields xmlns=\"URI\">>true</flag><string xmlns=BAZ>test</string><number xmlns=URI>1</number></StructWithNamedFields>"
```
djc, line 32: The attribute values need to be quoted.

instant-xml-macros/src/lib.rs
choinskib: Agree, make sense, I simplified it, in my opinion, looks better now.

instant-xml-macros/src/lib.rs
choinskib: Not relevant anymore, it was not needed, good catch though, thanks.

instant-xml-macros/src/lib.rs
choinskib: Not relevant anymore after refactoring.
I might be missing something but it seems to be hard to do as `Nested Meta` is a local variable, as this:
[            let nested = match attr.parse_meta() {](https://github.com/InstantDomain/instant-xml/blob/1b5a5a3d6a77568768f29bc64ffc472db7714515/instant-xml-macros/src/lib.rs#L25)
is returning the `Meta` enum.

instant-xml/tests/all.rs
choinskib: fixed

instant-xml-macros/src/lib.rs
```diff
@@ -2,62 +2,129 @@ extern crate proc_macro;
...
-        }
+const XML: &str = "xml";
+
+struct Serializer {
+    default_namespace: Option<String>,
+    other_namespaces: HashMap<String, String>,
+}
+
+impl<'a> Serializer {
+    pub fn init(attributes: &'a Vec<syn::Attribute>) -> Serializer {
```
djc, line 16: This would idiomatically be called `new()`.

instant-xml-macros/src/lib.rs
```diff
@@ -2,62 +2,129 @@ extern crate proc_macro;
...
+        let mut default_namespace: Option<String> = None;
+        let mut other_namespaces = HashMap::<String, String>::default();
```
djc, line 18: Here and on the line above, I don't think we need the type annotations?

choinskib:
@djc I rebuilt it a little, it looks clearer for me, please let me know what you think.

instant-xml-macros/src/lib.rs
```diff
@@ -2,62 +2,129 @@ extern crate proc_macro;
...
+            let list = match nested.first() {
+                Some(NestedMeta::Meta(Meta::List(list))) => list,
+                _ => todo!(),
+            };
 
-        let list = match nested.first() {
-            Some(NestedMeta::Meta(Meta::List(list))) => list,
-            _ => todo!(),
-        };
+            if let Some(ident) = list.path.get_ident() {
```
djc, line 36: For this case, it would be better to use a `match` block so that you can do an early return if `get_ident()` yields `None` (or if `ident != "namespace"`).

instant-xml/tests/all.rs
```diff
@@ -4,10 +4,12 @@ use instant_xml::{FromXml, ToXml};
 struct Unit;
 
 #[derive(Debug, Eq, PartialEq, ToXml)]
-#[xml(namespace("URI"))]
+#[xml(namespace("URI", bar = "BAZ", foo = "BAR"))]
 struct StructWithNamedFields {
     flag: bool,
+    #[xml(namespace("bar"))]
```
djc, line 10: I think you meant `xml(namespace(bar))` here, in order to reference the previously defined `BAZ` namespace through a prefix? In that case the resulting serialization should include `xmlns:bar="BAZ" in the `StructWithNamedFields` start tag, and yield `<bar:string>test</bar:string>` here.

instant-xml/tests/all.rs
```diff
@@ -4,10 +4,12 @@ use instant_xml::{FromXml, ToXml};
...
     string: String,
+    #[xml(namespace("typo"))]
```
djc, line 12: Same here, though I think we should probably yield an error on serialization (not while deriving, since it could be that a parent element defines it for us) if we encounter a namespace prefix that hasn't been defined.

choinskib: Agree, but I would do the error handling as a separate PR.

PR #6, opened 2022-06-30
Serialization of custom fields

choinskib:


instant-xml-macros/src/lib.rs
```diff
@@ -106,7 +110,7 @@ impl<'a> Serializer {
         };
 
         output.extend(
-            quote!(+ ">" + self.#field_value.to_string().as_str() + "</" + #prefix + #field_name + ">"),
+            quote!(+ ">" + self.#field_value.to_xml(Some(child_prefixes)).unwrap().as_str() + "</" + #prefix + #field_name + ">"),
```
djc, line 113: I think the use of `+` to concatenate strings isn't very idiomatic, and the performance implications aren't as clear as some of the alternatives. Would be good to do something different here.

instant-xml-macros/src/lib.rs
```diff
@@ -181,13 +185,47 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
 
     serializer.add_footer(&root_name, &mut output);
 
+    let current_prefixes: BTreeSet<&str> = serializer.get_keys_set();
     TokenStream::from(quote!(
         impl ToXml for #ident {
-            fn write_xml<W: ::std::fmt::Write>(&self, write: &mut W) -> Result<(), instant_xml::Error> {
-                write.write_str(&(#output))?;
+            fn write_xml<W: ::std::fmt::Write>(&self, write: &mut W, parent_prefixes: Option<&mut std::collections::BTreeSet<&str>>) -> Result<(), instant_xml::Error> {
```
djc, line 191: You're not getting line wrapping from rustfmt here because you're inside a macro, but would be good to make sure lines are still wrapped after 100 columns.

instant-xml-macros/src/lib.rs
```diff
@@ -156,10 +160,10 @@ impl<'a> Serializer {
 #[proc_macro_derive(ToXml, attributes(xml))]
 pub fn to_xml(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-
     let ident = &ast.ident;
     let root_name = ident.to_string();
     let mut output: proc_macro2::TokenStream = TokenStream::from(quote!("".to_owned())).into();
+    let mut missing_prefixes = BTreeSet::new();
```
djc, line 166: It's not obvious to me why we need to track missing prefixes here?

instant-xml-macros/src/lib.rs
```diff
@@ -156,10 +160,10 @@ impl<'a> Serializer {
 #[proc_macro_derive(ToXml, attributes(xml))]
 pub fn to_xml(input: TokenStream) -> TokenStream {
     let ast = parse_macro_input!(input as syn::DeriveInput);
-
     let ident = &ast.ident;
     let root_name = ident.to_string();
     let mut output: proc_macro2::TokenStream = TokenStream::from(quote!("".to_owned())).into();
```
djc, line 165: This way of creating a `proc_macro2::TokenStream` seems much more complex than needed. Also I think the code will look cleaner if you import `proc_macro2::TokenStream` as `TokenStream` and make explicitly qualify `proc_macro::TokenStream`.

instant-xml-macros/src/lib.rs
```diff
@@ -181,13 +185,47 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
 
     serializer.add_footer(&root_name, &mut output);
 
+    let current_prefixes: BTreeSet<&str> = serializer.get_keys_set();
```
djc, line 188: Shouldn't need the type annotation here.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,48 @@ pub use macros::{FromXml, ToXml};
...
+            ) -> Result<(), Error> {
+                Ok(())
+            }
+
+            fn to_xml(
+                &self,
+                parent_prefixes: Option<&mut BTreeSet<&str>>,
+            ) -> Result<String, Error> {
+                let mut out = self.to_string();
+                self.write_xml(&mut out, parent_prefixes)?;
```
djc, line 42: We should find a way to do this that doesn't involve allocating a separate string for `self.to_string()`.

instant-xml/tests/all.rs
```diff
@@ -27,8 +51,35 @@ fn struct_with_named_fields() {
...
         .unwrap(),
         "<StructWithNamedFields xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><flag>true</flag><bar:string>test</bar:string><number xmlns=\"typo\">1</number></StructWithNamedFields>"
     );
 }
+
+#[test]
+fn struct_with_custom_field() {
+    assert_eq!(
+        StructWithCustomField {
+            test: Nested {
```
djc, line 64: I think we want to ditch the `<test>` tag here and only keep `<Nested>`.

instant-xml/tests/all.rs
```diff
@@ -27,8 +51,35 @@ fn struct_with_named_fields() {
...
+        .unwrap(),
+        "<StructWithCustomField xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><test><Nested><bar:flag>true</bar:flag></Nested></test></StructWithCustomField>"
+    );
+}
+
+#[test]
+#[should_panic]
+fn struct_with_custom_field_wrong_prefix() {
+    assert_eq!(
+        StructWithCustomFieldWrongPrefix {
```
djc, line 78: No need for the `assert_eq!()` here if the `unwrap()` is going to panic.

djc (commented):
Here's a round of review on this first PR. I think I'll hold off on reviewing the other ones until we get this one merged.

choinskib:
> BTW, this seems to target the `extended-namespaces` branch but I don't see a PR for that branch?

It is already merged #4, I will rebase this branch to main later on.

djc:
> > BTW, this seems to target the `extended-namespaces` branch but I don't see a PR for that branch?
> 
> It is already merged #4, I will rebase this branch to main later on.

Okay, please rebase it before my next round of review. 👍

instant-xml-macros/src/lib.rs
```diff
@@ -54,6 +54,13 @@ impl<'a> Serializer {
         }
     }
 
+    fn get_keys_set(&self) -> BTreeSet<&str> {
```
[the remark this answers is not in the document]

instant-xml-macros/src/lib.rs

instant-xml/tests/all.rs
choinskib: done

instant-xml-macros/src/lib.rs
choinskib: It is not valid anymore, as I changed the "+" with `push_str`, so it is formatted earlier during macro processing, not inside.

instant-xml-macros/src/lib.rs
choinskib: Changed to `push_str`

instant-xml-macros/src/lib.rs
choinskib: Missing_prefix means that there is no matching prefix at current level and it needs to be looked in higher levels (parent, grand parent, etc).  Later in `fn to_xml` we are chacking if prefix exist in higher levels and currently panic if not. Missing prefixes are the ones difined on the fields, we don't need the "not missed" ones, because we can confirm on compile time that those exist. We only need "missed" to check in runtime.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,47 @@ pub use macros::{FromXml, ToXml};
...
 pub trait ToXml {
-    fn write_xml<W: fmt::Write>(&self, write: &mut W) -> Result<(), Error>;
+    fn write_xml<W: fmt::Write>(
+        &self,
+        write: &mut W,
+        parent_prefixes: Option<&mut BTreeSet<&str>>,
+    ) -> Result<(), Error>;
 
-    fn to_xml(&self) -> Result<String, Error> {
+    fn to_xml(&self, parent_prefixes: Option<&mut BTreeSet<&str>>) -> Result<String, Error> {
```
djc, line 19: I don't think this needs the `parent_prefixes` argument, that's only necessary for the "internal" `write_xml()` method.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,47 @@ pub use macros::{FromXml, ToXml};
 pub mod parse;
 
 pub trait ToXml {
-    fn write_xml<W: fmt::Write>(&self, write: &mut W) -> Result<(), Error>;
+    fn write_xml<W: fmt::Write>(
+        &self,
+        write: &mut W,
+        parent_prefixes: Option<&mut BTreeSet<&str>>,
```
djc, line 16: Let's wrap `prefixes` in an opaque `Serializer` type and make it required. `Serializer` should implement `Default` so that callers can just pass `Serializer::default()` (instead of `None`).

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,47 @@ pub use macros::{FromXml, ToXml};
...
+                Ok(())
+            }
+
+            fn to_xml(
+                &self,
+                parent_prefixes: Option<&mut BTreeSet<&str>>,
+            ) -> Result<String, Error> {
+                self.write_xml(&mut self.to_string(), parent_prefixes)?;
+                Ok(self.to_string())
+            }
```
djc, line 43: We don't need to provide `to_xml()` here, since the trait provides a default implementation.

instant-xml-macros/src/lib.rs
```diff
@@ -54,60 +54,122 @@ impl<'a> Serializer {
...
+        let mut field_name = field.ident.as_ref().unwrap().to_string();
         let field_value = field.ident.as_ref().unwrap();
+        let field_type = if let syn::Type::Path(v) = &field.ty {
+            v.path.get_ident()
+        } else {
+            todo!();
+        };
+
+        // For non-scalars ditch the field name
+        let is_scalar = Self::is_scalar(field_type.as_ref().unwrap().to_string().as_str());
```
djc, line 118: We should not be comparing to the type name here. Instead we should allow the type's `ToXml` implementation to decide whether it wants to use the field name. So I guess we need to pass the field name into `write_xml()` as part of the serializer state.

instant-xml-macros/src/lib.rs
djc: In instant-xml-macros we still have all this code inside the `quote!()` macro, so it doesn't get formatted as usual IIRC.

choinskib: Hmm, I did some tests and when I'm expanding the code with `cargo expand` it is formatted correctly. I will dive deeper into this.

instant-xml/src/lib.rs
choinskib: It is moved to `Serializer` struct

instant-xml/src/lib.rs
choinskib: I changed it a little bit, please take a look

instant-xml-macros/src/lib.rs
choinskib: done, please take a look, I rebuilt the whole thing, seems simpler now

djc (commented):
This is looking much better, some questions left!

instant-xml/src/lib.rs
```diff
@@ -10,12 +10,176 @@ pub use macros::{FromXml, ToXml};
...
+
+pub enum FieldAttribute<'xml> {
+    Prefix(&'xml str),
+    Namespace(&'xml str),
+}
+
+pub struct FieldData<'xml> {
+    pub field_name: &'xml str,
+    pub field_attribute: Option<FieldAttribute<'xml>>,
+    pub value: String,
```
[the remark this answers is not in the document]
choinskib: Right! Thanks for that 😄.

instant-xml/src/lib.rs
```diff
@@ -10,12 +10,176 @@ pub use macros::{FromXml, ToXml};
...
+        Ok(output)
+    }
+
+    fn serialize(
+        &self,
+        serializer: &mut Serializer,
+        field_data: Option<&mut FieldData>,
+    ) -> Result<(), Error>;
+
+    fn write_xml(&self, serializer: &mut Serializer, field_data: &FieldData) -> Result<(), Error> {
```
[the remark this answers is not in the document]
choinskib: I removed it, looks better now

instant-xml/src/lib.rs
```diff
@@ -10,12 +10,176 @@ pub use macros::{FromXml, ToXml};
...
+            output: &mut output,
+        };
+        self.serialize(&mut serializer, None)?;
+        Ok(output)
+    }
+
+    fn serialize(
+        &self,
+        serializer: &mut Serializer,
+        field_data: Option<&mut FieldData>,
```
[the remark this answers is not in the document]
choinskib: The idea is that you just modify the field_data if you want to customize the tag name, without the creating the new value.
In case of the custom type, by default it is taking the struct type as the xml tag (here)[https://github.com/InstantDomain/instant-xml/blob/349af40add28a0bf12142055355f3d1c754f32ec/instant-xml-macros/src/lib.rs#L219]
But if you would like to overide the serialize fn for some struct, you could just modify the `name` inside `FieldData` .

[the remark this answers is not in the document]
djc: Who would (or, where would you) modify the `field_data`? I'm not sure this idea makes sense. The `field_data` should be exclusively used by a parent type to pass some context into a child type.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,165 @@ pub use macros::{FromXml, ToXml};
...
+
+macro_rules! to_xml_for_number {
+    ($typ:ty) => {
+        impl ToXml for $typ {
+            fn serialize(
+                &self,
+                serializer: &mut Serializer,
+                field_data: Option<&mut FieldData>,
+            ) -> Result<(), Error> {
+                let field_data = match field_data {
```
djc, line 39: I think we should return an error in the case a scalar type gets called with `None` `FieldData`.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,165 @@ pub use macros::{FromXml, ToXml};
...
+        let field_data = match field_data {
+            Some(field_data) => field_data,
+            None => &FieldData {
+                field_name: "bool",
+                field_attribute: None,
+            },
+        };
+
+        serializer.add_open_tag(field_data);
+        serializer.output.push_str(value);
```
djc, line 86: Let's inline the `match` expression from above here.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,165 @@ pub use macros::{FromXml, ToXml};
...
     }
 }
 
+pub enum FieldAttribute<'xml> {
+    Prefix(&'xml str),
+    Namespace(&'xml str),
+}
+
+pub struct FieldData<'xml> {
+    pub field_name: &'xml str,
```
djc, line 168: Let's rename `FieldData` to `FieldContext` and rename the `field_` prefixes from the field names (since these are kinda redundant anyway).

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,165 @@ pub use macros::{FromXml, ToXml};
...
+                let field_data = match field_data {
+                    Some(field_data) => field_data,
+                    None => &FieldData {
+                        field_name: stringify!($typ),
+                        field_attribute: None,
+                    },
+                };
+
+                serializer.add_open_tag(field_data);
+                serializer.output.push_str(&self.to_string());
```
djc, line 48: Should use `write!()` or something to write into the output instead of allocating a `String`, then writing it into the output.

instant-xml/src/lib.rs
```diff
@@ -10,15 +10,165 @@ pub use macros::{FromXml, ToXml};
 pub mod parse;
 
 pub trait ToXml {
-    fn write_xml<W: fmt::Write>(&self, write: &mut W) -> Result<(), Error>;
-
     fn to_xml(&self) -> Result<String, Error> {
-        let mut out = String::new();
-        self.write_xml(&mut out)?;
-        Ok(out)
+        let mut parent_prefixes = BTreeSet::new();
```
djc, line 14: I think we should keep the `W: fmt::Write` argument and put it in the `Serializer`. How painful is it if the `Serializer` has a generic type argument `W` for this? My experience from Askama is that virtual calls at this level can lead to pretty painful performance.

instant-xml-macros/src/lib.rs
```diff
@@ -181,22 +211,47 @@ pub fn to_xml(input: TokenStream) -> TokenStream {
...
+
+                // Check if prefix exist
+                #(
+                    if serializer.parent_prefixes.get(#missing_prefixes).is_none() {
+                        return Err(instant_xml::Error::WrongPrefix);
+                    }
+                )*;
+
+                // Adding current prefixes
+                let mut to_remove: Vec<&str> = Vec::new();
```
djc, line 231: We statically now `current_prefixes` so I don't think we have to create a list with them in order to be able to remove them? The loop below could just replace `to_remove` with an array expansion of the current prefixes.

instant-xml/src/lib.rs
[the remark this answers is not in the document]
choinskib: Like this:
`
#[derive(ToXml)]
struct Foo {
    bar: Bar,
}

struct Bar {
    flag: bool,
}

impl ToXml for Bar {
       fn serialize(.., field_data: ..) -> (...) {
                 field_data.unwrap().name = "Something"; // field_name passed from parent is "bar"
                 ...
       }
}
`
But I don't have a strong opinion about that, we can change it to immutable.

instant-xml/src/lib.rs
choinskib: It was meant to be used for tuple structs, but I don't know if we are planning to support them.

instant-xml-macros/src/lib.rs
choinskib: I might not follow, but the `current_prefixe` may hide the `parent_prefix`. This is the reason for remove, you don't want to remove prefix that existed in parent, we only have to remove the added ones. If there is simpler way of doing that then great, but nothing came to my mind.

instant-xml/src/lib.rs
choinskib: I changed it, but I do not get the performance implications here, could you elaborate more about it? We can make a short call as well.

instant-xml/src/lib.rs
djc: We should probably support tuple structs, but not with scalars (directly) inside them.

instant-xml/src/lib.rs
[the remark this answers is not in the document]
djc: Let's make it immutable for now, can always revisit the decision later.

instant-xml/src/lib.rs
djc: If we have a generic `W: fmt::Write`, callers will monomorphize the implementations with a particular `fmt::Write` implementations. If we instead hold a `&mut dyn fmt::Write` in the `Serializer`, each call to write some data into the `fmt::Write` will have to go through the method table used for the `&dyn` fat pointer. Given that we're doing a lot of calls into that, the performance implications for that start to add up.

instant-xml-macros/src/lib.rs
djc: Okay, fair enough. I guess my main concern was allocating a `Vec` for this, I guess we could just create an array here instead (since we know the length of the array at codegen time).

choinskib: We don't, as we don't know which prefixes already exist in parent. We are checking it in the next line.

djc: Okay, never mind, we can optimize this later.

djc (approved):

PR #7, opened 2022-07-06
Simple deserializer

choinskib:


instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,289 @@
+use crate::{get_namespaces, retrieve_attr};
+use proc_macro::TokenStream;
+use proc_macro2::{Ident, Span};
+use quote::quote;
+
+pub struct Deserializer<'a> {
+    /// Original input.
+    pub input: &'a syn::DeriveInput,
+    pub fn_deserialize: proc_macro2::TokenStream,
```
djc, line 9: Let's do the same treatment for imports as on the serializer side, and make all these fields private.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,289 @@
...
+                                );
+                            } else {
+                                Self::process_element(
+                                    field,
+                                    index,
+                                    &mut elements_enum,
+                                    &mut elements_names,
+                                    &mut elem_type_match,
+                                    &mut declare_values,
+                                    &mut return_val,
```
djc, line 65: Should probably wrap these in a type.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn process_element(
+        field: &syn::Field,
+        index: usize,
+        elements_enum: &mut proc_macro2::TokenStream,
+        elements_names: &mut proc_macro2::TokenStream,
+        elem_type_match: &mut proc_macro2::TokenStream,
+        declare_values: &mut proc_macro2::TokenStream,
+        return_val: &mut proc_macro2::TokenStream,
+    ) {
+        let mut field_name = field.ident.as_ref().unwrap().to_string();
```
djc, line 202: Looks like there's a lot of duplicated code between `process_element()` and `process_attribute()`.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,289 @@
...
+        .into();
+
+        Deserializer {
+            input,
+            fn_deserialize,
+            fn_from_xml,
+        }
+    }
+
+    fn is_scalar(value: &str) -> bool {
```
djc, line 189: Should fix this like on the serializer side.

instant-xml/src/impls.rs
```diff
@@ -0,0 +1,28 @@
...
+impl<'de> Visitor<'de> for BoolVisitor {
+    type Value = bool;
+
+    fn visit_str<'a>(self, value: &str) -> Result<Self::Value, Error> {
+        Ok(FromStr::from_str(value).unwrap())
+    }
+}
+
+impl<'xml> FromXml<'xml> for bool {
+    fn deserialize<D>(deserializer: &mut D) -> Result<Self, Error>
```
djc, line 15: I don't think we need a generic `D` here?

instant-xml/src/lib.rs
```diff
@@ -53,7 +69,181 @@ to_xml_for_type!(i32);
 to_xml_for_type!(String);
 
 pub trait FromXml<'xml>: Sized {
-    fn from_xml(input: &str) -> Result<Self, Error>;
+    fn from_xml(_input: &str) -> Result<Self> {
+        unimplemented!();
+    }
+
+    fn deserialize<D>(deserializer: &mut D) -> Result<Self>
```
djc, line 76: Should not have both `deserialize()` and `deserialize_attr()`, that should part of the `deserializer` state.

instant-xml/src/lib.rs
```diff
@@ -53,7 +69,181 @@ to_xml_for_type!(i32);
...
+    }
+}
+
+pub trait AccessorXml<'xml> {
+    fn peek_next_tag(&mut self) -> Result<Option<XmlRecord>>;
+    fn verify_prefix(&self, prefix_to_verify: &str) -> bool;
+}
+
+pub struct Deserializer<'xml> {
+    pub iter: &'xml mut XmlParser<'xml>,
```
djc, line 141: This should probably own both `iter` (which should probably be called `parser`) and `prefixes`.

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,152 @@
...
-impl<'a> Parse for Option<Result<xmlparser::Token<'a>, xmlparser::Error>> {
-    fn element_start(self, ns: Option<&str>, tag: &str) -> Result<(), Error> {
-        match self {
-            Some(Ok(Token::ElementStart { prefix, local, .. })) => {
-                let prefix_ns = prefix.as_str();
-                let (has_prefix, expect_prefix) = (!prefix_ns.is_empty(), ns.is_some());
-                if has_prefix != expect_prefix {
-                    return dbg!(Err(Error::UnexpectedValue));
-                }
+use crate::{Error, Result};
```
djc, line 1: Please separate imports in three blocks separated by a newline

* `std` and/or `core`
* external crates (xmlparser)
* crate-internal symbols

djc (commented):
Initial round of review to get you started...

instant-xml-macros/src/de.rs
choinskib: I had it combined and it was supr unreadable, a lot of ifs to follow, it looks similar but it is using different variables names, `attributes` vs `elements`. I would keep it separated at least for now.

instant-xml-macros/src/de.rs
choinskib: There is different order here then in Serializer. 
This is producing something like:
```
        enum __Elements {
            __Value0,
            __Value2,
            __Ignore,
        }
        fn get_element(value: &str) -> __Elements {
            match value {
                "flag" => __Elements::__Value0,
                "Nested" => __Elements::__Value2,
                _ => __Elements::__Ignore,
            }
        }
```
This means that you are expecting the `Nested` tag that will be later deserialized to `Nested` struct. You can overide it when manually implementing the Deserialaizer. But to make it generic you need to assume somthing, or have something like `get_name()` which would return the default name, in Deserialize trait for each type. Assuming that non-scalars types by default have the struct name as tag name, looks clearer for me.

The other option would be the attribute like `rename` so you would not have to implement manually the whole thing, but still by default for non-scalars, i would keep the struct name as default.

instant-xml/src/impls.rs
choinskib: This would require to change it everywhere to fixed Deserializer including the Visitor trait, I don't know if we want to do this this way? I don't have a strong opinion here.

instant-xml-macros/src/de.rs
djc: Hmm, ok.

instant-xml-macros/src/de.rs
djc: Having a hard-coded list of type names simply won't work (for example, `Option` might instead be spelled as `core::option::Option`), we need to defer that to a trait implementation.

instant-xml/src/impls.rs
djc: Yes, serde abstracts over this because it implements deserialization across many different formats, but for this crate we're just doing XML.

instant-xml-macros/src/de.rs
choinskib: Good point, but we will still need this function `is_scalar`, as for scalar types we always want to use struct field name as a tag, not the default name from trait.

djc: Sure, we will need to interrogate the `FromXml` impl to see if it's a scalar.

instant-xml-macros/src/de.rs
choinskib: done

instant-xml-macros/src/de.rs
choinskib: done

instant-xml-macros/src/de.rs
choinskib: done, I added the function tag_name

instant-xml/src/lib.rs
choinskib: I changed name to parser. Those `prefixes` (now `namespaces`) are from struct definition, not the part of the parser. The `tag_attributes` could probably be moved to parser, I will take a fresh look at it on Monday.

instant-xml/src/parse.rs
choinskib: done

instant-xml/src/lib.rs
choinskib: done

instant-xml/src/impls.rs
choinskib: done

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,163 @@ pub struct FieldContext<'xml> {
...
+            None => Err(Error::UnexpectedEndOfStream),
+        }
+    }
+
+    fn process_open_tag(
+        &mut self,
+        name: &str,
+        namespaces: &HashMap<&'xml str, &'xml str>,
+    ) -> Result<()> {
+        if let Some(item) = self.parser.next() {
```
djc, line 325: Should do early returns here as well as for the non-`Open` case below.

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,163 @@ pub struct FieldContext<'xml> {
...
+
+        Ok(ret)
+    }
+
+    fn deserialize_bool<V>(&mut self, visitor: V) -> Result<V::Value>
+    where
+        V: Visitor<'xml>,
+    {
+        self.parser.next();
+        if let Some(item) = self.parser.next() {
```
djc, line 292: Invert the condition and return early.

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,163 @@ pub struct FieldContext<'xml> {
...
+    }
+
+    fn deserialize_attribute<V>(&mut self, visitor: V) -> Result<V::Value>
+    where
+        V: Visitor<'xml>,
+    {
+        visitor.visit_str(&self.pop_next_attribute_value()?)
+    }
+
+    fn pop_next_attribute_value(&mut self) -> Result<String> {
```
djc, line 313: Can just inline this in `deserialize_attribute()`, given that it has no other callers?

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,163 @@ pub struct FieldContext<'xml> {
...
+    FieldName,
+    Custom(&'xml str),
+}
+
 pub trait FromXml<'xml>: Sized {
-    fn from_xml(input: &str) -> Result<Self, Error>;
+    const TAG_NAME: XMLTagName<'xml>;
+
+    fn from_xml(_input: &str) -> Result<Self> {
+        unimplemented!();
```
djc, line 217: I think this should have a default implementation so that none of the implementers have to implement this?

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,163 @@ pub struct FieldContext<'xml> {
...
+    }
+
+    fn deserialize(deserializer: &mut Deserializer, kind: EntityType) -> Result<Self>;
```
djc, line 220: Is there a reason `kind` cannot be part of the `Deserializer`?

instant-xml/src/lib.rs
```diff
@@ -150,7 +169,7 @@ impl<'xml, W: std::fmt::Write> Serializer<'xml, W> {
         Ok(())
     }
 
-    fn add_close_tag(&mut self, field_context: &FieldContext) -> Result<(), Error> {
+    fn add_close_tag(&mut self, field_context: &FieldContext) -> Result<()> {
```
djc, line 172: I know that some crates like to alias `Result`, but I'd much prefer not doing so, as I find it just makes it harder to reason about the code.

instant-xml/src/lib.rs
```diff
@@ -5,12 +5,31 @@ use thiserror::Error;
 pub use xmlparser;
 
 pub use macros::{FromXml, ToXml};
+use parse::XmlParser;
 
+pub mod impls;
 #[doc(hidden)]
 pub mod parse;
 
+pub struct TagData {
```
djc, line 14: Pretty sure all the members here could be `&str` instead of `String`?

instant-xml-macros/src/se.rs
```diff
@@ -0,0 +1,107 @@
+use std::collections::{BTreeSet, HashMap};
+
+use proc_macro2::TokenStream;
+use quote::quote;
+
+use crate::{get_namespaces, retrieve_field_attribute, FieldAttribute};
+
+pub struct Serializer {
```
djc, line 8: I'd like to have all the moving code around (and renaming) in a separate PR so that this PR only contains the deserializer changes.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut declare_values = TokenStream::new();
+        let mut return_val = TokenStream::new();
+
+        match &input.data {
+            syn::Data::Struct(ref data) => {
+                match data.fields {
+                    syn::Fields::Named(ref fields) => {
+                        fields.named.iter().enumerate().for_each(|(index, field)| {
+                            if let Some(true) = retrieve_attr("attribute", &field.attrs) {
+                                Self::process_field(
```
djc, line 68: If the only difference between the two calls to `process_field()` here is the 3rd and 6th arguments, we should write the code so that this is obvious.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,271 @@
...
+            names_: &mut elements_names,
+            match_: &mut elem_type_match,
+        };
+
+        // Attributes
+        let mut attributes_enum = TokenStream::new();
+        let mut attributes_lets = TokenStream::new();
+        let mut attributes_names = TokenStream::new();
+        let mut attr_type_match = TokenStream::new();
+        let mut attributes_tokens = Tokens {
```
djc, line 51: Since `Tokens` does not seem to escape this scope, it's unclear to me why `Tokens` holds `&mut` references to its members instead of just taking ownership. I'm guessing you could probably `derive(Default)` on `Tokens` and then call `Tokens::default()` here, drastically reducing the amount of boilerplate here.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,271 @@
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::quote;
+
+use crate::{get_namespaces, retrieve_attr};
+
+struct Tokens<'a> {
+    enum_: &'a mut TokenStream,
+    consts_: &'a mut TokenStream,
```
djc, line 8: Why are all these names `_`-suffixed? I think that's only necessary for `enum` and `match`.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,271 @@
...
+            proc_macro::TokenStream::from(quote!(
+                const TAG_NAME: ::instant_xml::XMLTagName<'xml> = ::instant_xml::XMLTagName::Custom(#name);
+            ))
+            .into(),
+        );
+
+        Deserializer { fn_vec }
+    }
+
+    pub fn fn_vec(&self) -> &Vec<TokenStream> {
```
djc, line 198: Maybe `Deserializer` should implement `ToTokens` instead?

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,271 @@
...
+        tokens.consts_.extend(quote!(
+            const #const_field_name: &str = match #field_type::TAG_NAME {
+                ::instant_xml::XMLTagName::FieldName => #field_name,
+                ::instant_xml::XMLTagName::Custom(v) => v,
+            };
+        ));
+
+        if is_element {
+            tokens.names_.extend(quote!(
+                if( value == #const_field_name ) {
```
djc, line 231: No need to put parentheses around the condition here? (And below.)

instant-xml-macros/src/lib.rs
```diff
@@ -1,185 +1,123 @@
...
-                    if let Some(NestedMeta::Lit(Lit::Str(v))) = iter.next() {
-                        default_namespace = Some(v.value());
-                    }
+pub(crate) fn get_namespaces(
+    attributes: &Vec<syn::Attribute>,
+) -> (Option<String>, HashMap<String, String>) {
+    let mut default_namespace = None;
+    let mut other_namespaces = HashMap::default();
+
+    if let Some(list) = retrieve_attr_list("namespace", attributes) {
```
djc, line 28: Should do early returns and early continues here to reduce rightward drift.

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,154 @@
...
+use xmlparser::{ElementEnd, Token, Tokenizer};
+
+use crate::{Error, Result};
+pub use crate::{TagData, XmlRecord};
 
-                if local.as_str() != tag {
-                    return dbg!(Err(Error::UnexpectedValue));
+pub struct XmlParser<'a> {
+    stack: Vec<String>,
+    internal_iter: Peekable<Tokenizer<'a>>,
```
djc, line 11: I think we can just get rid of the `internal_` prefix here?

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,154 @@
...
-                }
+use xmlparser::{ElementEnd, Token, Tokenizer};
+
+use crate::{Error, Result};
+pub use crate::{TagData, XmlRecord};
 
-                if local.as_str() != tag {
-                    return dbg!(Err(Error::UnexpectedValue));
+pub struct XmlParser<'a> {
+    stack: Vec<String>,
```
djc, line 10: Could `stack` contain `&'a str` instead of `String`?

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,154 @@
...
+            let item = match self.internal_iter.next() {
+                Some(v) => v,
+                None => return Ok(None),
+            };
+
+            println!("{:?}", &item);
+            match item {
+                Ok(Token::ElementStart { prefix, local, .. }) => {
+                    key = local.to_string();
+                    prefix_ret = if prefix.is_empty() {
```
djc, line 39: I prefer to write these as `match` expressions, which I think are easier to read:

```rust
match prefix.is_empty() {
    true => None,
    false => Some(prefix.to_string()),
}
```

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,154 @@
...
 
-pub trait Parse {
-    fn element_start(self, ns: Option<&str>, tag: &str) -> Result<(), Error>;
-    fn element_end(self, ns: Option<&str>, tag: &str) -> Result<(), Error>;
+impl<'a> Iterator for XmlParser<'a> {
+    type Item = Result<XmlRecord>;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.parse_next() {
```
djc, line 148: Any reason not to just inline the implementation of `parse_next()` here?

instant-xml/tests/all.rs
```diff
@@ -9,7 +9,7 @@ struct StructWithCustomField {
     test: Nested,
 }
 
-#[derive(Debug, Eq, PartialEq, ToXml)]
+#[derive(Debug, Eq, PartialEq, ToXml, FromXml, Clone)]
```
djc, line 12: Why do we need `Clone` here?

instant-xml/tests/all.rs
```diff
@@ -37,10 +37,20 @@ struct StructWithNamedFields {
     number: i32,
 }
 
+#[derive(Debug, Eq, PartialEq, FromXml, ToXml)]
```
djc, line 40: Let's keep this definition close to the test.

instant-xml/tests/all.rs
```diff
@@ -81,3 +91,104 @@ fn struct_with_custom_field_wrong_prefix() {
     .to_xml()
     .unwrap();
 }
+
+#[test]
+fn struct_with_custom_field_from_xml() {
+    assert_eq!(
+        StructWithCustomFieldFromXml::from_xml("<StructWithCustomFieldFromXml flag_attribute=\"true\" xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><bar:flag>false</bar:flag><Nested><flag>true</flag></Nested></StructWithCustomFieldFromXml>").unwrap(),
```
djc, line 98: BTW, I think I missed this on the serializer side, but this is actually wrong due to namespaces.

Because the XML here sets the default namespace to `URI`, the `Nested` tag must be interpreted as `{URI}baz`, so deserialization should fail in this case because the `Nested` type does not declare its namespace as `URI`.

instant-xml/tests/all.rs
choinskib: Some leftover, removed

instant-xml/src/parse.rs
choinskib: Inlined

instant-xml/src/lib.rs
choinskib: Done

instant-xml/src/lib.rs
choinskib: done

instant-xml-macros/src/se.rs
choinskib: I only did this, because Serializer and Deserializer are using the same functions, that were erlier only in Serializer.

djc: I'm not objecting to the change itself, just makes it harder for me to review the actually changed code if they're in the same PR.

instant-xml/src/lib.rs
choinskib: Please take a look now, I simplified it.

instant-xml/src/lib.rs
choinskib: Please take a look now, I simplified it.

instant-xml-macros/src/de.rs
choinskib: I changed the arguments order so now the last ones varies. I hope this is what you meant.

instant-xml-macros/src/de.rs
choinskib: It is more about passing the `Tokens` into the `quote!`, because `Tokens` do not implements the ToTokens, and I need separete variables so I will not change it into stream. As far as I understand, even if `Tokens` would keep the values, I would need to assign them to `let some_token :TokenStream`, it seems as boilerplaty as current solution.  I migth by missing something though.

djc: Yeah, I guess we need bindings to `Tokens`'s members in order to splice them into the token stream. However, would still help not to have to repeat all the `TokenStream::new()` initialization.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,264 @@
...
+            #(#vec)*
+        ));
+    }
+}
+
+impl Deserializer {
+    pub fn new(input: &syn::DeriveInput) -> Deserializer {
+        let ident = &input.ident;
+        let name = ident.to_string();
+        let mut fn_vec = Vec::new();
```
djc, line 30: It looks to me like `fn_vec` is just the output token stream. Maybe it should be called `out` or something and be of type `TokenStream`? You can append a `TokenStream` to another by using `TokenStream::extend()` IIRC.

instant-xml-macros/src/de.rs
djc: No, I meant that there should only be one call to `Self::process_field()`. You can then switch the arguments around where the argument values need to be different using an `if`.

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,151 @@ pub struct FieldContext<'xml> {
...
+                        return Err(Error::MissingdPrefix);
+                    }
+                }
+
+                self.tag_attributes = item.attributes;
+                return Ok(());
+            }
+        }
+
+        Err(Error::UnexpectedValue)
```
djc, line 334: Untested, but this should be more like this:

```suggestion
        let item = match self.parser.next() {
            Some(Ok(XmlRecord::Open(item))) if item.key == name => item,
            _ => return Err(Error::UnexpectedValue),
        };

        for (k, v) in item.namespaces.unwrap() {
            let item = match item.namespaces.get(k) {
                Some(item) if *item == v => {
                     self.tag_attributes = item.attributes;
                     return Ok(());
                }
                Some(item) => return Err(Error::UnexpectedPrefix),
                None => return Err(MissingPrefix),
            }
```

instant-xml/src/parse.rs
```diff
@@ -1,48 +1,142 @@
...
+                    key = Some(local.as_str());
+                    prefix_ret = match prefix.is_empty() {
+                        true => None,
+                        false => Some(prefix.as_str()),
+                    };
+                }
+                Ok(Token::ElementEnd { end, .. }) => match end {
+                    ElementEnd::Open => {
+                        self.stack.push(key.unwrap());
+                        println!(
```
djc, line 89: Should remove all the `println!()` stuff before merging.

instant-xml-macros/src/de.rs
choinskib: I unified it, but it requires some overhead, if we are deciding inside of function, we don't know which `tokens` we need to append to. Please take a look at this commit:
https://github.com/InstantDomain/instant-xml/pull/7/commits/5966cde0a06b14ed7619c9ba1dca9164436a691f

instant-xml-macros/src/de.rs
choinskib: done

instant-xml-macros/src/lib.rs

instant-xml-macros/src/de.rs
choinskib: done

instant-xml/src/lib.rs
choinskib: Mostly because I wanted to avoid state-like processing, please take a look at this commit: https://github.com/InstantDomain/instant-xml/pull/7/commits/266443baeff1b2a61a7007bf5312b6b6622e1eda
Maybe you have some better idea of how to approch this?

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,289 @@
...
+            syn::Data::Struct(ref data) => {
+                match data.fields {
+                    syn::Fields::Named(ref fields) => {
+                        fields.named.iter().enumerate().for_each(|(index, field)| {
+                            match Self::process_field(
+                                field,
+                                index,
+                                &mut declare_values,
+                                &mut return_val,
+                            ) {
```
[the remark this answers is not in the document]
choinskib: done

instant-xml/src/lib.rs
choinskib: done

instant-xml-macros/src/de.rs
choinskib: done

instant-xml/tests/all.rs
choinskib: This along with other identified issues, will be addressed in this PR: #10

instant-xml-macros/src/se.rs
choinskib: Opened PR: #11

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+pub struct Deserializer {
+    out: TokenStream,
+}
+
+impl quote::ToTokens for Deserializer {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let vec = &self.out;
```
djc, line 30: I think we can do something like `tokens.extend([self.out.clone()])`?

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+impl Deserializer {
+    pub fn new(input: &syn::DeriveInput) -> Deserializer {
+        let ident = &input.ident;
+        let name = ident.to_string();
+        let mut out = TokenStream::new();
+
+        let (_, other_namespaces) = get_namespaces(&input.attrs);
+        let mut namespaces_map: TokenStream = proc_macro::TokenStream::from(
```
djc, line 44: The use of `proc_macro::TokenStream` here should not be needed.

instant-xml/src/lib.rs
```diff
@@ -181,8 +200,168 @@ pub struct FieldContext<'xml> {
     pub attribute: Option<FieldAttribute<'xml>>,
 }
 
+#[derive(Clone, PartialEq, Eq)]
+pub enum EntityType {
+    Element,
+    Attribute,
+}
+
+pub enum XMLTagName<'xml> {
```
djc, line 209: Let's just name this `TagName`.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,266 @@
...
+        field: &syn::Field,
+        index: usize,
+        declare_values: &mut TokenStream,
+        return_val: &mut TokenStream,
+        tokens: &mut Tokens,
+        is_element: bool,
+    ) {
+        let field_name = field.ident.as_ref().unwrap().to_string();
+        let const_field_name = Ident::new(&field_name.to_uppercase(), Span::call_site());
+        let field_value = field.ident.as_ref().unwrap();
```
djc, line 203: We probably don't need `field_value` since we already have `field_name`, which amounts to the same thing? Should factor out the common `.as_ref().unwrap()` at least.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,266 @@
...
+        let field_type = if let syn::Type::Path(v) = &field.ty {
```
djc, line 204: Better to spell this as a `match`.

instant-xml-macros/src/de.rs
```diff
@@ -0,0 +1,266 @@
...
+        ));
+
+        if is_element {
+            tokens.match_.extend(quote!(
+                __Elements::#enum_name => {
+                    if #enum_name.is_some() {
+                        panic!("duplicated value");
+                    }
+
+                    if item.prefix.is_some() {
```
djc, line 241: `is_some()` then `unwrap()` is an anti-pattern.

instant-xml-macros/src/lib.rs
```diff
@@ -1,185 +1,120 @@
...
 
 const XML: &str = "xml";
 
-enum FieldAttribute {
+pub(crate) enum FieldAttribute {
     Namespace(String),
     PrefixIdentifier(String),
 }
 
-struct Serializer {
```
djc, line 15: This diff should not be part of this PR anymore, right? Otherwise you merged in a strange way.

instant-xml/src/impls.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use crate::{Deserializer, EntityType, Error, FromXml, Visitor, XMLTagName};
+
+struct BoolVisitor;
+
+impl<'de> Visitor<'de> for BoolVisitor {
+    type Value = bool;
+
+    fn visit_str<'a>(self, value: &str) -> Result<Self::Value, Error> {
+        Ok(FromStr::from_str(value).unwrap())
```
djc, line 11: Should not `unwrap()` here but yield `Error` instead.

instant-xml-macros/src/lib.rs
choinskib: This because you merged it back 😄, the #11 PR was pointing to this branch not `main`.

instant-xml-macros/src/de.rs
choinskib: We use both of them in `quote!`, so I believe we need both here, unless there is other way to pass it?

djc: I just mean we should factor out the common part. And the naming seems unclear given that both have the same value?

instant-xml-macros/src/lib.rs
djc: Ahh, so you can you do one pointing at main? Then merge main into this branch after we merge it.

choinskib: Sure, it's here: #12

instant-xml-macros/src/de.rs
choinskib: Changed to field_var and field_var_str, if you have some better naming idea, please let me know.

instant-xml/src/parse.rs
djc: So time to do this?

choinskib: I believe so, I put a hold on this, in case you have some additional comments.

djc: Nah, I think this is in a decent shape, time to move on!

choinskib: done

djc (approved):

PR #10, opened 2022-08-11
Namespaces fixes - deserializer

choinskib:
This PR will address the identified issues in namespace handling. #7

### DIFF
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -33,24 +33,24 @@ impl Deserializer {
 
         // Elements
         let mut elements_enum = TokenStream::new();
-        let mut elements_lets = TokenStream::new();
+        let mut elements_consts = TokenStream::new();
         let mut elements_names = TokenStream::new();
         let mut elem_type_match = TokenStream::new();
         let mut elements_tokens = Tokens {
             enum_: &mut elements_enum,
-            consts_: &mut elements_lets,
+            consts_: &mut elements_consts,
             names_: &mut elements_names,
             match_: &mut elem_type_match,
         };
 
         // Attributes
         let mut attributes_enum = TokenStream::new();
-        let mut attributes_lets = TokenStream::new();
+        let mut attributes_consts = TokenStream::new();
         let mut attributes_names = TokenStream::new();
         let mut attr_type_match = TokenStream::new();
         let mut attributes_tokens = Tokens {
             enum_: &mut attributes_enum,
-            consts_: &mut attributes_lets,
+            consts_: &mut attributes_consts,
             names_: &mut attributes_names,
             match_: &mut attr_type_match,
         };
@@ -114,9 +114,11 @@ impl Deserializer {
                 }
 
                 fn get_element(value: &str) -> __Elements {
-                    #elements_lets
-                    #elements_names
-                    __Elements::__Ignore
+                    #elements_consts
+                    match value {
+                        #elements_names
+                        _ => __Elements::__Ignore
+                    }
                 }
 
                 enum __Attributes {
@@ -125,9 +127,11 @@ impl Deserializer {
                 }
 
                 fn get_attribute(value: &str) -> __Attributes {
-                    #attributes_lets
-                    #attributes_names
-                    __Attributes::__Ignore
+                    #attributes_consts
+                    match value {
+                        #attributes_names
+                        _ => __Attributes::__Ignore
+                    }
                 }
 
                 struct StructVisitor;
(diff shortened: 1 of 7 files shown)


instant-xml/tests/all.rs
```diff
@@ -61,7 +62,7 @@ fn struct_with_custom_field() {
         }
         .to_xml()
         .unwrap(),
-        "<StructWithCustomField xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><Nested><bar:flag>true</bar:flag></Nested></StructWithCustomField>"
+        "<StructWithCustomField xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><Nested xmlns=\"URI\" xmlns:bar=\"BAZ\"><bar:flag>true</bar:flag></Nested></StructWithCustomField>"
```
djc, line 65: In this case we should not repeat the `xmlns=` and `xmlns:bar=` attributes for the `Nested` tag, because they are implied by the parent tag.

If you like, we can save this for the future since this serialization is correct, but in that case I'd like you to file an issue for it so what we don't forget.

instant-xml/tests/all.rs
```diff
@@ -104,16 +105,223 @@ fn struct_with_custom_field_from_xml() {
...
+        Error::UnexpectedPrefix
+    );
+
+    // Other namespace not-nested - wrong parser namespace
+    assert_eq!(
+        NestedOtherNamespace::from_xml(
+            "<NestedOtherNamespace xmlns=\"URI\" xmlns:bar=\"WRONG\"><bar:flag>true</bar:flag></NestedOtherNamespace>"
+        )
+        .unwrap_err(),
+        Error::UnexpectedPrefix
```
djc, line 244: I think we should yield `Error::WrongNamespace` in all these cases, since prefixes are just syntactic sugar for the actual underlying namespaces.

instant-xml-macros/src/lib.rs
```diff
@@ -50,66 +52,46 @@ pub(crate) fn get_namespaces(
...
+        },
+        Some((None, name)) if name.as_str() == "attribute" => Some(FieldAttribute::Attribute),
+        None => None,
+        _ => panic!("unexpected parameter"),
     }
-    None
 }
 
-pub(crate) fn retrieve_attr(name: &str, attributes: &Vec<syn::Attribute>) -> Option<bool> {
+fn retrieve_attr_list(attributes: &Vec<syn::Attribute>) -> Option<(Option<syn::MetaList>, String)> {
```
djc, line 74: I don't think the values returned here needed to be owned?

instant-xml-macros/src/de.rs
```diff
@@ -58,16 +59,26 @@ impl Deserializer {
...
+                                Some(FieldAttribute::Namespace(value)) => {
+                                    field_namespace = Some(value);
+                                    (&mut elements_tokens, None, true)
                                 }
-                                _ => {
-                                    is_element = true;
-                                    &mut elements_tokens
+                                Some(FieldAttribute::PrefixIdentifier(def_prefix)) => {
+                                    if other_namespaces.get(&def_prefix).is_none() {
+                                        panic!("Namespace with such prefix do not exist for this struct");
```
djc, line 70: So this should yield an error, right?

instant-xml/tests/all.rs
choinskib: You are right, this PR name is a little misleading, as it evolved from simple-deserializer so it is touching only the deserializer part. I'm currently working on the Serializer part, I was thinking of separating it into other PR, are you ok with it? This change was just the convenience for the test to pass as `Nested` struct is used in some other tests, I can separate better though.

djc: Sure, it's fine to separate it into a separate PR.

instant-xml-macros/src/de.rs
choinskib: Actually, this happens in compile time, that's why it is `panic!`

instant-xml-macros/src/lib.rs
choinskib: `parse_meta` is returning the value, so we can not return the reference for those values here.

instant-xml/tests/all.rs
choinskib: I separated the serializer test and will open the serializer PR soon.

instant-xml-macros/src/lib.rs
djc: Ah, fair.

djc (approved):

PR #13, opened 2022-08-26
Serializer namespaces and attributes

choinskib:


instant-xml-macros/src/lib.rs
```diff
@@ -33,19 +33,22 @@ pub(crate) fn get_namespaces(
 
     if name == "namespace" {
         let mut iter = list.nested.iter();
-        if let Some(NestedMeta::Lit(Lit::Str(v))) = iter.next() {
+        let mut next = iter.next();
+        if let Some(NestedMeta::Lit(Lit::Str(v))) = next {
```
djc, line 37: What problem is this trying to solve? The iterator state is in `iter` so it should not be necessary to keep `next` as a separate binding here.

instant-xml-macros/src/se.rs
```diff
@@ -67,42 +70,79 @@ impl<'a> Serializer {
         ));
     }
 
-    pub fn process_named_field(
-        &mut self,
-        field: &syn::Field,
-        output: &'a mut TokenStream,
-        missing_prefixes: &'a mut BTreeSet<String>,
-    ) {
+    pub fn process_named_field(&mut self, field: &syn::Field) -> (TokenStream, bool) {
```
djc, line 73: We should pass in a `&mut TokenStream` instead of allocating a separate one in here.

instant-xml-macros/src/se.rs
```diff
@@ -67,42 +70,79 @@ impl<'a> Serializer {
...
-            self.#field_value.serialize(serializer, Some(&field))?;
+            serializer.set_field_context(field)?;
+            self.#field_value.serialize(serializer)?;
+            serializer.retrive_parent_default_namespace();
         ));
+
+        (output, is_attribute)
+    }
+
+    pub fn get_namespaces_token(&self) -> TokenStream {
```
djc, line 132: No `get_()` prefixes please.

instant-xml-macros/src/se.rs
```diff
@@ -67,42 +70,79 @@ impl<'a> Serializer {
...
+        let mut namespaces = quote!(
+            let mut to_remove: Vec<&str> = Vec::new();
+        );
+        for (k, v) in self.other_namespaces.iter() {
+            namespaces.extend(quote!(
+                // Only adding to HashMap if namespace do not exist, if it exist it will use the parent defined prefix
+                if let std::collections::hash_map::Entry::Vacant(v) = serializer.parent_namespaces.entry(#v) {
+                        v.insert(#k);
```
djc, line 140: You have a bunch of 8-column indents in this PR, please clean up the indentation (since `fmt` doesn't handle it inside the `quote!()` macro calls).

instant-xml/src/lib.rs
```diff
@@ -185,11 +256,14 @@ impl<'xml, W: std::fmt::Write> Serializer<'xml, W> {
...
 }
 
+#[derive(Clone)]
 pub enum FieldAttribute<'xml> {
     Prefix(&'xml str),
     Namespace(&'xml str),
+    Attribute,
 }
 
+#[derive(Clone)]
```
djc, line 266: Why do we need this?

instant-xml/tests/all.rs
```diff
@@ -1,28 +1,17 @@
...
-}
+//TODO: Add compile time errors check?
 
 #[derive(Debug, Eq, PartialEq, ToXml)]
 struct Unit;
 
 #[test]
 fn unit() {
-    assert_eq!(Unit.to_xml().unwrap(), "<Unit></Unit>");
+    assert_eq!(Unit.to_xml().unwrap(), "<Unit xmlns=\"\"></Unit>");
```
djc, line 9: We should only a serialize an element's default namespace if it's different from a parent's element default namespace (which is not the case here because there is no parent element).

instant-xml/tests/all.rs
```diff
@@ -41,52 +35,104 @@ fn struct_with_named_fields() {
...
+    flag_internal_prefix: bool,
+}
+
 #[derive(Debug, Eq, PartialEq, ToXml)]
 #[xml(namespace("URI", bar = "BAZ", foo = "BAR"))]
 struct StructWithCustomField {
+    #[xml(attribute)]
+    int_attribute: i32,
+    #[xml(namespace("BAZ"))]
+    flag_direct_namespace_same_as_prefix: bool,
```
djc, line 57: For this case, I think we should serialize ` xmlns="BAZ"` instead of using the prefix.

We're slightly overloading the meaning of the unnamed first namespace argument to mean both the namespace that an element is in *and* the default namespace for that element. These aren't quite the same thing, and what you have currently is correct, but I think it would be more consistent for now to use `BAR` as the default in this case.

instant-xml/tests/all.rs
```diff
@@ -298,7 +344,7 @@ fn other_namespaces() {
 #[derive(Debug, Eq, PartialEq, FromXml)]
 #[xml(namespace("URI"))]
 struct StructDirectNamespace {
-    #[xml(namespace("BAZ"))]
+    #[xml(namespace("BAZ"))] // TODO: Zmienić to na URI i zobaczyć czy działa
```
djc, line 347: This isn't very helpful for me. 😄

instant-xml/tests/all.rs
```diff
@@ -41,52 +35,104 @@ fn struct_with_named_fields() {
...
+        }
+        .to_xml()
+        .unwrap(),
+        "<StructChildNamespaces xmlns=\"URI\" xmlns:bar=\"BAZ\" xmlns:foo=\"BAR\"><NestedDifferentNamespace xmlns=\"\" xmlns:internal=\"INTERNAL\"><bar:flag_parent_prefix>true</bar:flag_parent_prefix><internal:flag_internal_prefix>false</internal:flag_internal_prefix></NestedDifferentNamespace><Nested xmlns:internal=\"INTERNAL\"><bar:flag_parent_prefix>true</bar:flag_parent_prefix><internal:flag_internal_prefix>false</internal:flag_internal_prefix></Nested></StructChildNamespaces>"
+    );
+}
+
+#[derive(Debug, Eq, PartialEq, FromXml)]
+#[xml(namespace("URI", bar = "BAZ"))]
+struct NestedDe {
```
djc, line 133: This is unused, get rid of it?

instant-xml-macros/src/lib.rs
choinskib: It is about default namespace, to be able to parse something like:
`#[xml(namespace(bar = "BAZ", foo = "BAR"))]`
and 
`#[xml(namespace("BAZ", foo = "BAR"))]`
but not
`#[xml(namespace( foo = "BAR",  "BAZ"))]`
or
`#[xml(namespace("BAZ", "BAR")]`

djc: I was referring to the particular change of keeping a separate `next` binding instead of referring to `iter.next()` directly.

instant-xml/tests/all.rs
choinskib: This is used, there had to be some mismatch with tests at some point. It is used in the tests for deserializer.

instant-xml-macros/src/se.rs
choinskib: done

instant-xml-macros/src/se.rs
choinskib: done

instant-xml-macros/src/se.rs
choinskib: done

instant-xml/src/lib.rs
choinskib: Here: https://github.com/InstantDomain/instant-xml/blob/022eed647090451e416ffec6b5b16880701759a5/instant-xml/src/lib.rs#L188

instant-xml/tests/all.rs
choinskib: done

instant-xml/tests/all.rs
choinskib: removed

instant-xml/src/lib.rs
djc: Okay, you can use `take()` instead.

instant-xml/src/lib.rs
```diff
@@ -185,7 +185,7 @@ impl<'xml, W: std::fmt::Write> Serializer<'xml, W> {
     }
 
     pub fn consume_field_context(&mut self) -> Option<FieldContext<'xml>> {
-        let ret = self.next_field_contest.clone();
+        let ret = self.next_field_contest.take();
```
djc, line 188: You can just do `self.next_field_contest.take()` (by the way, should rename that `contest` -> `context`), it will leave the field as `None`.

choinskib: done

djc (approved):

PR #14, opened 2022-08-27
Deserializer scalars

### LINKED ISSUE #16: Handle entities
XML has 5 pre-defined character entities which we should be able to encode and decode. These are:

* `&` -> `&amp;`
* `"` => `&quot;`
* `<` => `&lt;`
* `>` => `&gt;`
* `'` => `&apos;`

Thus, when serializing a `&str` or `String` or similar, if any of these characters occurs, we should replace them in the output. When deserializing, there are typically three possible target types:

* `&str`: we should make sure the input string does not contain entities, and yield an error (or panic? I forget what serde does, but I guess we should copy its behavior) if it does
* `String`: decode the entity before initializing the output type
* `Cow<_, 'str>`: yield a `Cow::Borrowed` if the input string does not contain entities, `Cow::Owned` if it does

(Additionally a DTD may define custom entities in the XML header/prolog. This is something to keep in mind for the design but let's skip this feature for now -- as I understand it, it's not heavily used.)

choinskib:
Close #16


instant-xml-macros/src/de.rs
```diff
@@ -34,6 +34,25 @@ impl quote::ToTokens for Deserializer {
 impl Deserializer {
     pub fn new(input: &syn::DeriveInput) -> Deserializer {
         let ident = &input.ident;
+        let generics = (&input.generics).into_token_stream();
```
djc, line 37: Probably want to use `split_for_impl()` here, too. I recently wrote something similar like this:

```rust
    let mut email_generics = ast.generics.clone();
    let email_lifetime = syn::LifetimeDef::new(syn::Lifetime::new("'email", Span::call_site()));
    for lt in email_generics.lifetimes_mut() {
        lt.bounds.push(email_lifetime.lifetime.clone());
    }
    email_generics
        .params
        .push(syn::GenericParam::Lifetime(email_lifetime));

    // Split the generics for use in impls and type definitions, below.

    let (_, inner_ty_generics, _) = ast.generics.split_for_impl();
    let (impl_generics, ty_generics, where_clause) = email_generics.split_for_impl();
```

(You may not need both of the split sets.)

instant-xml/src/impls.rs
```diff
@@ -20,8 +23,128 @@ impl<'xml> FromXml<'xml> for bool {
...
+struct CharVisitor;
+impl<'xml> Visitor<'xml> for CharVisitor {
+    type Value = char;
+
+    fn visit_str(self, value: &str) -> Result<Self::Value, Error> {
+        if value.len() != 1 {
+            return Err(Error::Other("Expected char type".to_string()));
+        }
+
+        Ok(value.chars().next().expect("char type"))
```
djc, line 117:
```suggestion
        match value.len() {
            1 => Ok(value.chars().next().expect("char type")),
            _ => Err(Error::Other("Expected char type".to_string())),
        }
```

instant-xml/src/impls.rs
```diff
@@ -20,8 +23,128 @@ impl<'xml> FromXml<'xml> for bool {
...
+impl<'xml> FromXml<'xml> for &'xml str {
+    const TAG_NAME: TagName<'xml> = TagName::FieldName;
+
+    fn deserialize(deserializer: &mut Deserializer<'xml>) -> Result<Self, Error> {
+        match deserializer.consume_next_type() {
+            EntityType::Element => deserializer.deserialize_element(StrVisitor),
+            EntityType::Attribute => deserializer.deserialize_attribute(StrVisitor),
+        }
+    }
+}
```
djc, line 150: Here, too, let's add a `Cow<'xml, str>` impl.

choinskib: Done

instant-xml/src/impls.rs
choinskib: done

instant-xml-macros/src/de.rs
choinskib: Unfortunately, it is not very helpful here, at the end we would need something similar to what serde is doing [here](https://github.com/serde-rs/serde/blob/f52d134c14f6904010d0a59107987b050b36b811/serde_derive/src/bound.rs). But it will take more time to implement, so I would leave it as it is for now.

djc: Why is it not helpful here?

choinskib: Because what I can get useful from `split_for_impl` is basically this:` let generics = (&input.generics).into_token_stream();`
What I need is: 
e.g.
": 'a + 'b, 'a, 'b" and ", 'a, 'b"

This is not about getting the generics from the struct is about combining them with our 'xml lifetime to have something like:
` struct StructVisitor<'xml: 'a + 'b, 'a, 'b> {
            marker: std::marker::PhantomData<StructDeserializerScalars<'a, 'b>>,
            lifetime: std::marker::PhantomData<&'xml ()>,
        }
        impl<'xml: 'a + 'b, 'a, 'b> Visitor<'xml> for StructVisitor<'xml, 'a, 'b> {`
        Unless you are thinking about something I don't get?

choinskib: Still, don't get how to get something like`<'xml: 'a + 'b, 'a, 'b>` from it. I will try to look at it once more later.

djc: I definitely think we need `split_for_impl()` here, but before we do that we probably need to modify the original `Generics` to add the `'xml` lifetime.

instant-xml-macros/src/de.rs
```diff
@@ -207,15 +246,56 @@ impl Deserializer {
...
+                Some(ident) => ident.into_token_stream(),
+                None => {
+                    if v.path.segments.is_empty() {
+                        panic!("Wrong declaration");
+                    };
+
+                    // Cow<>, Option<>
+                    let mut out =
+                        (&v.path.segments.first().expect("type name").ident).into_token_stream();
+                    match &(v.path.segments.first().expect("type name")).arguments {
```
djc, line 259: So what problem is this trying to solve?

instant-xml/src/impls.rs
```diff
@@ -20,8 +24,165 @@ impl<'xml> FromXml<'xml> for bool {
...
     fn deserialize(deserializer: &mut Deserializer) -> Result<Self, Error> {
         match deserializer.consume_next_type() {
-            EntityType::Element => deserializer.deserialize_bool(BoolVisitor),
+            EntityType::Element => deserializer.deserialize_element(BoolVisitor),
             EntityType::Attribute => deserializer.deserialize_attribute(BoolVisitor),
         }
     }
 }
+
+struct NumberVisitor<T>
```
djc, line 33: Similar to what I mentioned on the serializer side, I think we should have a helper type here that can handle all of these conversions which can basically rely on `FromStr`.

instant-xml/src/impls.rs
```diff
@@ -20,8 +24,165 @@ impl<'xml> FromXml<'xml> for bool {
...
+from_xml_for_number!(isize);
+from_xml_for_number!(u8);
+from_xml_for_number!(u16);
+from_xml_for_number!(u32);
+from_xml_for_number!(u64);
+from_xml_for_number!(usize);
+from_xml_for_number!(f32);
+from_xml_for_number!(f64);
+
+struct StringVisitor;
```
djc, line 88: Please add newlines after these type definitions.

instant-xml-macros/src/de.rs
djc: I guess my question is, why do we need to do that?

instant-xml/src/impls.rs
choinskib: done

instant-xml/src/impls.rs
choinskib: done

instant-xml-macros/src/de.rs
djc: After some experimentation, I couldn't get the lifetimes to work. I rewrote the code to discard lifetimes to be a little more robust, seems to be working for now.

instant-xml-macros/src/de.rs
djc: Will take care of this separately.

djc (approved):

PR #19, opened 2022-08-29
Option type for deserializer

choinskib:
Leaving it as a draft for now, as #14 should go first. Some refactoring might also be needed here.

### DIFF
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -190,7 +190,6 @@ impl Deserializer {
                                     }
                                  }
                                  XmlRecord::Close(tag) => {
-                                    println!("Close: {}", tag);
                                     if tag == &#name {
                                         break;
                                     }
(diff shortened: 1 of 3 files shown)


instant-xml-macros/src/de.rs
```diff
@@ -242,13 +241,49 @@ impl Deserializer {
...
+                                        }
+                                    }
+                                    _ => panic!("Wrong data"),
+                                }
+
+                                temp.extend(quote!(>));
+                                temp
+                            }
+                            _ => panic!("Wrong data"),
+                        },
```
djc, line 283: I definitely think this is wrong, and we should find another way to do this. I've been looking at the serde output and source code but haven't quite figured it out yet.

instant-xml-macros/src/de.rs
```diff
@@ -278,9 +313,15 @@ impl Deserializer {
...
-        declare_values.extend(quote!(
-            let mut #enum_name: Option<#field_type> = None;
-        ));
+        if is_option_type {
+            declare_values.extend(quote!(
+                let mut #enum_name: Option<#field_type> = Some(None);
+            ));
+        } else {
+            declare_values.extend(quote!(
+                let mut #enum_name: Option<#field_type> = None;
```
djc, line 322: serde definitely just sets this to `None` for `Option` values, too.

instant-xml-macros/src/de.rs
choinskib: This looks ugly, because I need to manually remove lifetime from the definition, otherwise it would be much easier.
Something like: `v.path.segments.first().expect("Struct name").into_tokens_stream()`. Serde is using this function heavily, maybe it will be helpful: https://github.com/serde-rs/serde/blob/f52d134c14f6904010d0a59107987b050b36b811/serde_derive/src/de.rs#L3120
I will take a look. But I will focus first on the rest of PRs.

instant-xml-macros/src/de.rs
choinskib: Yes, they have this function that handles that:
https://github.com/serde-rs/serde/blob/f52d134c14f6904010d0a59107987b050b36b811/serde/src/private/de.rs#L20

choinskib:
Moved into #14

PR #24, opened 2022-09-21
Serialize scalar enums

### LINKED ISSUE #18: Allow using unit-variant enums as scalars
I'd like to be able to define an enum **that only has unit variants** to be serialized/deserialized as a scalar.

For example:

```rust
#[xml(scalar)]
enum Foo {
    Bar,
    Baz,
}
```

Should serialize as the string `Bar`. Deserializing any other string should yield an error. Depending on the order with #17, we should also support using `rename_all` on enums to rename variants. While you're in there, in the case that the variant has an assigned value we should serialize that value:

```rust
#[xml(scalar)]
enum Foo {
    Bar = 12,
    Baz,
}
```

It should be a compile-time error to have `#[xml(scalar)]` on an enum that has non-unit variants like `Foo(u8)` or `Bar { n: usize }`.

rsdy:
This patchset resolves #18 .

The proc macro will panic if:
 * The enum is not marked as `scalar` with `#[xml(scalar)]`
 * The variant has a non-literal discriminant (e.g. `1 + 1`, or some other `const` expr)
 * The `#[xml(scalar)]` attribute is used for variants (only permitted on the enum)
 * A variant has both a discriminant and a `#[xml(rename = "value")]` attribute.

### DIFF
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -5,7 +5,20 @@ use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace};
 
 pub(crate) fn from_xml(input: &syn::DeriveInput) -> TokenStream {
     let ident = &input.ident;
-    let container_meta = ContainerMeta::from_derive(input);
+    let meta = ContainerMeta::from_derive(input);
+
+    match &input.data {
+        syn::Data::Struct(ref data) => deserialize_struct(input, data, meta, ident),
+        _ => todo!(),
+    }
+}
+
+fn deserialize_struct(
+    input: &syn::DeriveInput,
+    data: &syn::DataStruct,
+    container_meta: ContainerMeta,
+    ident: &Ident,
+) -> TokenStream {
     let default_namespace = match &container_meta.ns.uri {
         Some(ns) => quote!(#ns),
         None => quote!(""),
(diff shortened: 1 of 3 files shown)


instant-xml-macros/src/de.rs
```diff
@@ -1,11 +1,54 @@
...
-use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace};
+use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace, VariantMeta};
 
 pub(crate) fn from_xml(input: &syn::DeriveInput) -> TokenStream {
     let ident = &input.ident;
     let container_meta = ContainerMeta::from_derive(input);
+    match &input.data {
+        syn::Data::Struct(ref data) => deserialize_struct(input, data, container_meta, ident),
+        syn::Data::Enum(_) if !container_meta.scalar => {
+            panic!("non-scalar enums are currently unsupported!")
```
djc, line 12: I'd like to have all new compile-time errors generate a `syn::Error` (similar to what `ser::process_named_field()` is already doing).

instant-xml-macros/src/de.rs
```diff
@@ -1,11 +1,54 @@
...
 use quote::{quote, ToTokens};
 
-use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace};
+use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace, VariantMeta};
 
 pub(crate) fn from_xml(input: &syn::DeriveInput) -> TokenStream {
     let ident = &input.ident;
     let container_meta = ContainerMeta::from_derive(input);
+    match &input.data {
+        syn::Data::Struct(ref data) => deserialize_struct(input, data, container_meta, ident),
```
djc, line 10: For now, I'd like `Struct` variants to yield an error if `container_meta.scalar` is `true`.

Also, please separate out the change to extract a `deserialize_struct()` function.

instant-xml-macros/src/de.rs
```diff
@@ -1,11 +1,54 @@
...
+        syn::Data::Enum(_) if !container_meta.scalar => {
+            panic!("non-scalar enums are currently unsupported!")
+        }
+        syn::Data::Enum(ref data) => deserialize_enum(input, data),
+        _ => todo!(),
+    }
+}
+
+#[rustfmt::skip]
```
djc, line 19: Why do we need the `rustfmt::skip` here?

instant-xml-macros/src/lib.rs
```diff
@@ -59,12 +61,66 @@ impl FieldMeta {
...
+        if discriminant.is_some() && rename.is_some() {
+            panic!("conflicting `rename` attribute and variant discriminant!")
+        }
+
+        let serialize_as = rename
+            .or(discriminant)
+            .map(|lit| lit.into_token_stream())
+            .unwrap_or_else(|| input.ident.to_string().to_token_stream());
+
+        VariantMeta { serialize_as }
```
djc, line 120: I'd prefer to inline the `serialize_as` value here as a `match rename.or(discriminant) { }` (instead of using combinators on both sides).

instant-xml-macros/src/ser.rs
```diff
@@ -4,25 +4,66 @@ use syn::spanned::Spanned;
...
-            match data.fields {
-                syn::Fields::Named(ref fields) => {
-                    fields.named.iter().for_each(|field| {
-                        process_named_field(field, &mut body, &mut attributes, &meta);
-                    });
-                }
-                syn::Fields::Unnamed(_) => todo!(),
-                syn::Fields::Unit => {}
-            };
+        syn::Data::Struct(ref data) => serialize_struct(input, data, meta),
```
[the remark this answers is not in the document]
rsdy: See https://github.com/InstantDomain/instant-xml/pull/24/commits/1349ba7e8405a823a20acbad3242f11d496e80bf

instant-xml-macros/src/de.rs
rsdy: Generally rustfmt misbehaves on code inside macros, so anything with a macro invocation where I want to preserve formatting, I'll use `rustfmt::skip`.

djc:
So I'd like the third commit squashed into the second one, since it's not logically separate.

On rustfmt vs macros, can you try and see what actually changes about the formatting when you remove the `#[skip]` attributes? So far it's been fine in this crate (and in my experience on other crates).

PR #25, opened 2022-09-23
`rename_all` container attribute

### LINKED ISSUE #17: Allow user to rename fields and types
serde has a number of ways to allow users to rename fields and types. On `struct` types, we want to enable:

* `#[xml(rename = "foo")] struct Foo {}`, which would replace the use of `Foo` with `foo`
* `#[xml(rename_all = "PascalCase")] struct Foo {}`, which would rename all fields in `struct` according to the named convention

For fields, we just want to support `#[xml(rename = "foo")]`. Remember that these can compose, so you might encounter `#[xml(rename = "foo", namespace("URI"))]`.

We'll want to use the heck crate for these conversions. Except for `Title Case`, I think we want to support all four of heck's support variants.

rsdy:
Enable a container attribute `rename_all` that uses the renamer rules from `serde`. The following syntax is permitted to rename *attributes* of a struct or *variants* of an enum. These changes will **not** rename the XML tag generated from the  struct name itself.

```rust
#[xml(scalar, rename_all = "UPPERCASE")]
pub enum TestEnum {
    Foo = 1,
    Bar,
    Baz
}

#[xml(rename_all = "UPPERCASE")]
pub struct TestStruct {
    field_1: String,
    field_2: u8,
}
```

The `rename_all` attribute may take on the following values:

 *  `lowercase`
 *  `UPPERCASE`
 *  `PascalCase`
 *  `camelCase`
 *  `snake_case`
 *  `SCREAMING_SNAKE_CASE`
 *  `kebab-case`
 *  `SCREAMING-KEBAB-CASE`

This PR resolves #17 .

### DIFF
--- a/instant-xml-macros/src/case.rs
+++ b/instant-xml-macros/src/case.rs
@@ -11,7 +11,7 @@ use std::fmt::{self, Debug, Display};
 use self::RenameRule::*;
 
 /// The different possible ways to change case of fields in a struct, or variants in an enum.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum RenameRule {
     /// Don't apply a default rename rule.
     None,
--- a/instant-xml-macros/src/case.rs
+++ b/instant-xml-macros/src/case.rs
@@ -11,7 +11,7 @@ use std::fmt::{self, Debug, Display};
 use self::RenameRule::*;
 
 /// The different possible ways to change case of fields in a struct, or variants in an enum.
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum RenameRule {
     /// Don't apply a default rename rule.
     None,
@@ -34,6 +34,12 @@ pub enum RenameRule {
     KebabCase,
     /// Rename direct children to "SCREAMING-KEBAB-CASE" style.
     ScreamingKebabCase,
+}
+
+impl Default for RenameRule {
+    fn default() -> Self {
+        None
+    }
 }
 
 static RENAME_RULES: &[(&str, RenameRule)] = &[
--- a/instant-xml-macros/src/case.rs
+++ b/instant-xml-macros/src/case.rs
@@ -1,0 +1,198 @@
+//! Originally from https://raw.githubusercontent.com/serde-rs/serde/master/serde_derive/src/internals/case.rs
+//! Code to convert the Rust-styled field/variant (e.g. `my_field`, `MyType`) to the
+//! case of the source (e.g. `my-field`, `MY_FIELD`).
+
+// See https://users.rust-lang.org/t/psa-dealing-with-warning-unused-import-std-ascii-asciiext-in-today-s-nightly/13726
+#[allow(deprecated, unused_imports)]
+use std::ascii::AsciiExt;
+
+use std::fmt::{self, Debug, Display};
+
+use self::RenameRule::*;
+
+/// The different possible ways to change case of fields in a struct, or variants in an enum.
+#[derive(Copy, Clone, PartialEq)]
+pub enum RenameRule {
+    /// Don't apply a default rename rule.
+    None,
+    /// Rename direct children to "lowercase" style.
+    LowerCase,
+    /// Rename direct children to "UPPERCASE" style.
+    UpperCase,
+    /// Rename direct children to "PascalCase" style, as typically used for
+    /// enum variants.
+    PascalCase,
+    /// Rename direct children to "camelCase" style.
+    CamelCase,
+    /// Rename direct children to "snake_case" style, as commonly used for
+    /// fields.
+    SnakeCase,
+    /// Rename direct children to "SCREAMING_SNAKE_CASE" style, as commonly
+    /// used for constants.
+    ScreamingSnakeCase,
+    /// Rename direct children to "kebab-case" style.
+    KebabCase,
+    /// Rename direct children to "SCREAMING-KEBAB-CASE" style.
+    ScreamingKebabCase,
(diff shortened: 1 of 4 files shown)


djc:
#17 requested the use of heck, why did you decide to not do that?

rsdy:
Since serde's usecase is equivalent to this, the boilerplate provided by the `case.rs` file allows us to straightforwardly plug in `heck` to do the conversion if need be. I don't see the reason to use the library at this stage, though, since this is already a comprehensive and well-tested suite.

As for the string literals instead of paths, that's an easy change to make.

instant-xml-macros/src/case.rs
```diff
@@ -0,0 +1,204 @@
...
+    ScreamingKebabCase,
+}
+
+impl Default for RenameRule {
+    fn default() -> Self {
+        None
+    }
+}
+
+static RENAME_RULES: &[(&str, RenameRule)] = &[
```
djc, line 45: Can this be `const` instead?

instant-xml-macros/src/lib.rs
```diff
@@ -40,6 +44,12 @@ impl ContainerMeta {
                 MetaItem::Attribute => panic!("attribute key invalid in container xml attribute"),
                 MetaItem::Ns(ns) => meta.ns = ns,
                 MetaItem::Rename(lit) => meta.rename = Some(lit),
+                MetaItem::RenameAll(lit) => {
+                    meta.rename_all = match RenameRule::from_str(&lit.to_string()) {
+                        Ok(rule) => rule,
+                        Err(err) => panic!("{err}"),
```
djc, line 50: This should use a `syn::Error`, and we should use the `lit` so that we can preserve the span. (As such, I don't think we need the `ParseError` to carry the unknown name?)

instant-xml-macros/src/lib.rs
```diff
@@ -55,17 +65,29 @@ struct FieldMeta {
...
         let mut meta = FieldMeta::default();
         for item in meta_items(&input.attrs) {
             match item {
                 MetaItem::Attribute => meta.attribute = true,
                 MetaItem::Ns(ns) => meta.ns = ns,
                 MetaItem::Rename(lit) => meta.rename = Some(lit),
-                MetaItem::Scalar => panic!("attribute 'scalar' invalid in field xml attribute"),
+                MetaItem::RenameAll(_) => {
+                    return Err(syn::Error::new(
+                        input.span(),
```
djc, line 77: We should use the `item.span()` here.

instant-xml-macros/src/lib.rs
```diff
@@ -55,17 +65,29 @@ struct FieldMeta {
...
+                        "attribute 'rename_all' invalid in field xml attribute",
+                    ))
+                }
+                MetaItem::Scalar => {
+                    return Err(syn::Error::new(
+                        input.span(),
```
djc, line 83: And here.

instant-xml-macros/src/lib.rs
```diff
@@ -98,7 +125,12 @@ impl VariantMeta {
                         "attribute 'ns' is invalid for enum variants",
                     ))
                 }
-                MetaItem::Rename(lit) => rename = Some(lit.to_token_stream()),
+                MetaItem::RenameAll(_) => {
+                    return Err(syn::Error::new(
+                        input.span(),
```
djc, line 130: Should use the `item.span()` here.

instant-xml-macros/src/case.rs
```diff
@@ -0,0 +1,204 @@
...
+                return Ok(*rule);
+            }
+        }
+        Err(ParseError {
+            unknown: rename_all_str,
+        })
+    }
+
+    /// Apply a renaming rule to an enum variant, returning the version expected in the source.
+    pub fn apply_to_variant(&self, variant: &str) -> String {
```
djc, line 69: I think `apply_to_variant()` and `apply_to_field()` should take a `&Ident` so that we can yield the `ident.to_string()` allocation in case no change is needed.

instant-xml-macros/src/lib.rs
```diff
@@ -454,10 +491,17 @@ fn meta_items(attrs: &[syn::Attribute]) -> Vec<MetaItem> {
             (MetaState::Rename, TokenTree::Punct(punct)) if punct.as_char() == '=' => {
                 MetaState::RenameValue
             }
+            (MetaState::RenameAll, TokenTree::Punct(punct)) if punct.as_char() == '=' => {
```
djc, line 494: Please move this arm below the `RenameValue, Literal` arm.

instant-xml-macros/src/lib.rs
rsdy: We can `span()` on the `Literal` that comes through, but not on the `item` `MetaItem` instance, which is disconnected from the `syn` types that implement `Spanned`.

instant-xml-macros/src/lib.rs
rsdy: See above, I think we should be consistent with how reporting happens.

instant-xml-macros/src/lib.rs
rsdy: Same as above.

instant-xml-macros/src/lib.rs
djc: Okay, I guess doing it on the `Literal` wouldn't be obviously better. Can we do the `input.attrs` instead?

rsdy: I think the way to do this would be to emit a `Vec<(MetaItem, Span)>` from the `meta_items` function.

djc: Yeah, but let's leave that for a separate PR.

djc (approved):

PR #26, opened 2022-09-26
Better error reporting

rsdy:
This PR retrieves the span for every parsed attribute on a field or container.

This allows bubbling this information up to the call site and generating a specific error message instead of using a broader span or `panic!` macro.


djc:
Can you rebase on top of main?

djc:
This is good. To bad `Span::join()` is not yet stable.

PR #27, opened 2022-09-26
Add integration test for `rename_all` attribute

rsdy:
Simple test cases to ensure fields and enum variants are all renamed correctly using the `rename_all` macro attribute.

### DIFF
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{quote, ToTokens};
-use syn::spanned::Spanned;
+use syn::{spanned::Spanned, ImplGenerics};
 
 use super::{discard_lifetimes, ContainerMeta, FieldMeta, Namespace, VariantMeta};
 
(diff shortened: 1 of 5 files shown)


rsdy:
Writing the test for `enum`s immediately popped out an error in the generated code. Will fix that here.

rsdy:
The test for `enum`s also fixes the enum serialization code. Ready for review.

instant-xml-macros/src/de.rs
```diff
@@ -11,22 +11,53 @@ pub(crate) fn from_xml(input: &syn::DeriveInput) -> TokenStream {
         Err(e) => return e.to_compile_error(),
     };
 
+    let name = match &meta.rename {
```
djc, line 14: I would like to split this refactoring off into a separate commit.

instant-xml-macros/src/ser.rs
```diff
@@ -56,6 +61,8 @@ fn serialize_enum(
             ) -> Result<(), instant_xml::Error> {
 		serializer.write_str(match self { #variants })
             }
+
+            const KIND: ::instant_xml::Kind = ::instant_xml::Kind::Scalar;
```
djc, line 65: This is arguably a separate change, which also merits its own commit.

instant-xml/tests/rename.rs
```diff
@@ -23,7 +23,7 @@ fn renamed() {
 }
 
 #[test]
-fn rename_all() {
+fn rename_all_struct() {
```
djc, line 26: Please squash this into the first commit.

instant-xml-macros/src/ser.rs
rsdy: This is part of fixing the serialization logic, though. This constant is part of the trait.

instant-xml-macros/src/de.rs
rsdy: I can move `default_namespace` and `name` back to the `deserialize_struct` function, as ultimately they're not needed here, so the minimum changeset becomes cleaner.

As a separate PR it would probably make sense to split these into a struct and define `(de)?serialize_{enum,struct}` functions on that parameter struct. WDYT?

instant-xml-macros/src/ser.rs
djc: I guess it would be okay to have the required fixes in the same commit as the additional test(s).

djc (approved):

PR #28, opened 2022-09-27
Reduce code redundancy

rsdy:
Extract code into a common source that are used between ser/de and struct/enum functions.

I think this PR hits the inflection point of diminishing returns. There's probably some space for further improvements, but it would require quite a major refactor of certain components, which are not worth it right now.

### DIFF
--- a/instant-xml-macros/src/lib.rs
+++ b/instant-xml-macros/src/lib.rs
@@ -8,11 +8,11 @@ use std::collections::BTreeMap;
 use std::fmt;
 
 use proc_macro2::{Delimiter, Group, Ident, Literal, Punct, Span, TokenStream, TokenTree};
-use quote::ToTokens;
-use syn::parse_macro_input;
+use quote::{quote, ToTokens};
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
 use syn::token::Colon2;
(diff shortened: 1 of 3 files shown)


instant-xml-macros/src/ser.rs
```diff
@@ -67,6 +67,10 @@ fn serialize_struct(
     data: &syn::DataStruct,
     meta: ContainerMeta,
 ) -> proc_macro2::TokenStream {
+    let ident = &input.ident;
+    let tag = meta.tag();
```
djc, line 71: Would prefer to bind `tag` and `default_namespace` as close to their usage site as makes sense (similar to where they were before) -- same for all the users.

Why do we need to take a reference of `meta.default_namespace()`?

instant-xml-macros/src/lib.rs
```diff
@@ -47,18 +52,49 @@ impl ContainerMeta {
...
+        Ok(Self {
+            input,
+            ns,
+            rename,
+            rename_all,
+            scalar,
+        })
+    }
+
+    fn lifetimed_generics(&self) -> Generics {
```
djc, line 76: Using lifetime as a verb feels a little odd. Maybe just call it `xml_generics()`?

djc (commented):
Nice!

PR #29, opened 2022-09-28
Add support for `Vec<T>`

rsdy:
This PR adds support for `Vec<T>`, and will produce the following XML for a `Vec<String>`:

    <list>
    <element>one</element>
    <element>two</element>
    <element>three</element>
    </list>


instant-xml/tests/entities.rs
```diff
@@ -10,32 +10,34 @@ struct StructSpecialEntities<'a> {
...
     cow: Cow<'a, str>,
+    vec: Vec<String>,
 }
 
 #[test]
 fn escape_back() {
     assert_eq!(
         from_str(
-            "<StructSpecialEntities xmlns=\"URI\"><string>&lt;&gt;&amp;&quot;&apos;adsad&quot;</string><str>str</str><cow>str&amp;</cow></StructSpecialEntities>"
+            "<StructSpecialEntities xmlns=\"URI\"><string>&lt;&gt;&amp;&quot;&apos;adsad&quot;</string><str>str</str><cow>str&amp;</cow><list xmlns=\"\"><element xmlns=\"\">one</element><element xmlns=\"\">two</element><element xmlns=\"\">three</element></list></StructSpecialEntities>"
```
djc, line 20: I think we should use the field name and namespace as the list element tag (so `vec` in this case). I think `Vec`s will need a third variant in `Kind`. We should only have the `element` tag in case `T` is `Kind::Scalar`, otherwise we should use the nested type's element directly.

We definitely need tests for both a scalar `T` (which you already have) and a non-scalar one.

djc (approved):
Definitely more sticky edge cases than expected, but I think this is a pretty good first take!

PR #35, opened 2023-02-21
Check character boundaries before comparing string slices

djc:


valkum (commented):
Wouldn't the line with `result.push_str(input.get(last_end..last_end + 1).unwrap());`  (533) also be susceptible to larger UTF-8 characters?

djc:
@valkum good point! I decided to invest a little, and rewrote the `decode()` implementation with an explicit state machine that is a little shorter, quite a bit faster and more likely to be correct -- comes with benchmarks and tests.

instant-xml/src/impls.rs
```diff
@@ -769,3 +770,23 @@ impl<'xml> FromXml<'xml> for IpAddr {
...
+        assert_eq!(decode("foo &lt; bar").unwrap(), "foo < bar");
+        assert_eq!(decode("foo &gt; bar").unwrap(), "foo > bar");
+        assert_eq!(decode("foo &quot; bar").unwrap(), "foo \" bar");
+        assert_eq!(decode("foo &apos; bar").unwrap(), "foo ' bar");
+        assert_eq!(decode("foo &amp;lt; bar").unwrap(), "foo &lt; bar");
+        assert_eq!(decode("&amp; foo").unwrap(), "& foo");
+        assert_eq!(decode("foo &amp;").unwrap(), "foo &");
+        assert!(decode("&foo;").is_err());
+        assert!(decode("&foobar;").is_err());
+    }
```
valkum, line 791: Let's add some more utf8 values here to be safe.
* `cbdtéda&amp;sü` -> `cbdtéda&sü`
* `cbdtéd&ampü` -> should error I guess

djc: Cool, one of those actually found a bug. Changed `decode()` to work in the byte domain, which simplifies the non-error path code and seems to be slightly faster too.

valkum (approved):
LGTM. Maybe we want to do some fuzzing in the future.

PR #36, opened 2023-02-22
Remaining changes to support epp-client port (in particular, inline mode)

djc:
Inline mode can be used for (de)serializing a group of elements together, without wrapping it in a shared parent element. This is necessary for EPP extensions, where we sometimes need to deal with multiple extensions inside an `<extension>` tag.

@valkum I can imagine you don't want to do a very detailed review of this, but please look at the test cases and see if things conceptually make sense to you. These are the remaining changes needed to get the instant-epp tests to pass.

### DIFF
--- a/instant-xml-macros/src/case.rs
+++ b/instant-xml-macros/src/case.rs
@@ -95,7 +95,11 @@ impl RenameRule {
 
     /// Apply a renaming rule to a struct field, returning the version expected in the source.
     pub fn apply_to_field(&self, ident: &Ident) -> String {
-        let field = ident.to_string();
+        let mut field = ident.to_string();
+        if field.starts_with("r#") {
+            field = field[2..].to_string();
+        }
+
         match *self {
             None | LowerCase | SnakeCase => field,
             UpperCase => field.to_ascii_uppercase(),
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -67,8 +67,9 @@ fn deserialize_scalar_enum(
             }
 
             fn deserialize<'cx>(
-                deserializer: &mut ::instant_xml::Deserializer<'cx, 'xml>,
                 into: &mut Self::Accumulator,
+                field: &'static str,
+                deserializer: &mut ::instant_xml::Deserializer<'cx, 'xml>,
             ) -> Result<(), ::instant_xml::Error> {
                 use ::instant_xml::Error;
 
@@ -104,6 +105,7 @@ fn deserialize_forward_enum(
     }
 
     let ident = &input.ident;
+    let field_str = format!("{ident}::0");
     let mut matches = TokenStream::new();
     let mut variants = TokenStream::new();
     let mut borrowed = BTreeSet::new();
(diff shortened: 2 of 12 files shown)


instant-xml/tests/inline.rs
```diff
@@ -0,0 +1,39 @@
...
+struct Wrapper {
+    inline: Inline,
+}
+
+#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[xml(inline)]
+struct Inline {
+    foo: Foo,
+    bar: Bar,
+}
```
valkum, line 13: Does this also support the following?
```rust
struct Wrapper {
    inline: Vec<Inline>,
}

#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
#[xml(inline)]
enum Inline {
    Foo(Foo),
    Bar(Bar),
}
```
Wondering which data structure is better. Imagine Vec to be some non-allocating collection.

instant-xml/tests/inline.rs
```diff
@@ -0,0 +1,39 @@
...
+    let v = Wrapper {
+        inline: Inline {
+            foo: Foo { i: 42 },
+            bar: Bar {
+                s: "hello".to_string(),
+            },
+        },
+    };
+
+    let xml = r#"<Wrapper><Foo><i>42</i></Foo><Bar><s>hello</s></Bar></Wrapper>"#;
```
valkum, line 36: Maybe also a negative test for this data structure?
```xml
<Wrapper><Foo><i>42</i><Bar><s>hello</s></Bar></Foo></Wrapper>
```

valkum (commented):
Just a small question and an additional test case for now. 
Took some time to understand what inline mode means. Maybe `transparent` or `synthetic` is a better name? We also should invest some time into docs and examples.

instant-xml/tests/inline.rs
djc: Yeah, I think that should work? Don't have a use case for it yet. The case in instant-epp where we need this is where we're explicitly combining two extensions so in that case IMO being explicit about the two fields is nicer.

instant-xml/tests/inline.rs
djc: Good suggestion, added.

djc:
> Maybe `transparent` or `synthetic` is a better name?

Renamed it to `transparent` to mirror serde.

> We also should invest some time into docs and examples.

Yup, that's next on my list.

instant-xml/tests/inline.rs
valkum: My use case was also part of an idea for building your own set of supported extensions, e.g. like this:
```rust
// In lib
struct Wrapper<T = Inline> {
    inline: Vec<T>,
}

#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
enum Inline {
    Foo(Foo),
    Bar(Bar),
}

// In your own code you can then do something like this
type Response = Wrapper<AllExtensionsEnabledResponse>;
enum AllExtensionsEnabledResponse {
    Foo(Foo),
    Bar(Bar),
    Baz(Baz)
}
```
Which one is more ergonomic probably depends on the exact case (e.g. if all fields are optional or not)
Downside would be one more generic inside. Could hurt compile times.

valkum (approved):

instant-xml/tests/inline.rs
djc: Sure, but for now it seems like the current approach will do? At least I'm not sure what bearing this thread has on this PR.

PR #39, opened 2023-02-28
Add support for Cow<'_, [T]>

djc:
Would like to use this in instant-epp.

### DIFF
--- a/instant-xml/src/impls.rs
+++ b/instant-xml/src/impls.rs
@@ -319,11 +319,7 @@ impl<'xml> FromXml<'xml> for &'xml str {
     const KIND: Kind = Kind::Scalar;
 }
 
-impl<'xml, 'a, T: ?Sized> FromXml<'xml> for Cow<'a, T>
-where
-    T: ToOwned,
-    T::Owned: FromXml<'xml>,
-{
+impl<'xml, 'a> FromXml<'xml> for Cow<'a, str> {
     #[inline]
     fn matches(id: Id<'_>, field: Option<Id<'_>>) -> bool {
         match field {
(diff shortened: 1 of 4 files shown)


instant-xml/tests/scalar.rs
```diff
@@ -71,6 +73,7 @@ fn scalars() {
             },
             cow: Cow::from("123"),
             option: Some("asd"),
+            slice: [1, 2, 3][..].into(),
```
valkum, line 76: As we are talking about `Cows` here, might it make sense to explicitly define the cow here? Calling `into()` pretty much hides if the resulting `Cow` is owned or borrowed.

instant-xml/src/impls.rs
```diff
@@ -334,23 +330,93 @@ where
...
-    type Accumulator = Option<Cow<'a, T>>;
+    type Accumulator = CowSliceAccumulator<T>;
     const KIND: Kind = Kind::Scalar;
 }
 
+pub struct CowSliceAccumulator<T>
+where
+    [T]: ToOwned,
+{
+    pub(crate) inner: Vec<T>,
```
valkum, line 399: I would expect, based on the name, that this would be a `Cow<'a, [T]>`. Maybe this should be named `VecAccumulator` if it only contains owned types?

instant-xml/tests/scalar.rs
djc: Made these explicit.

instant-xml/src/impls.rs
valkum: Oh I think I get it now. This is needed to allow deserialization of the following, right?
```rs
struct A<'a, T>{
  field: Cow<T, 'a>
}
```
While deserializing, the matching occurrences are accumulated in a Vec (using `CowSliceAccumulator`)
and then, while constructing the `A` struct, are turned back into a Cow again (by calling `try_done`)?

I guess this is ok for now. But I think as soon as we have another newtype over `Vec` to implement Accumulate<T> for the newtype, we should unify them and rename this to `VecAccumulator`.

valkum (approved):

instant-xml/src/impls.rs
djc: Ohh, fair point. I changed it to use `Vec<T>` as the `Accumulator` for `Cow<'a, [T]>` and added a generic `Accumulate` impl to allow accumulating from a `Vec<T>` into the `Cow<'a, [T]>`.

PR #43, opened 2023-03-02
Implement ToXml for str instead of &str

djc:


valkum (commented):
It might be helpful to state the reason for the change in the commit. Is this to avoid a double ref in the serialize method of the impl, or for some downstream issues?

djc:
Yeah, it's to avoid the double ref. I ran into this while writing a manual `ToXml` impl in instant-epp.

valkum (approved):

PR #47, opened 2023-09-15
Correctly handle multi-byte characters in string encoding

djc:
Discovered a bug in encoding multi-byte characters that I ran into while working on RSS feeds. Cleaned up some other warnings that popped up since I last worked on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /instant-xml/target
 /instant-xml-macros/target
 Cargo.lock
+**/.DS_Store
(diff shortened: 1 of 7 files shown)


valkum (approved):
The multi-byte fix looks good.

Regarding `enum_name` -> `val_name` change: I assume we only need a name for the accumulator, and it can be anything, but should be unique, and lowercase? It doesn't need to be bound to the `enum_name` except via the match arm in `tokens.r#match`? If that's right, this looks good as well under the assumption that `index` is unique. `index` comes from `fields.named.iter().enumerate()`, and I think `Punctuated<T,P>::iter()` is stable. So it seems ok.

djc:
> Regarding enum_name -> val_name change: I assume we only need a name for the accumulator, and it can be anything, but should be unique, and lowercase?

Yup!

djc:
Published 0.3.2.

PR #55, opened 2024-04-01
Enable namespace prefixes containing dashes

### LINKED ISSUE #54: Namespace prefixes should be NCName
I was playing with this crate and came across an issue parsing namespace prefixes that contain `-`. Currently, it's not possible but according to the [spec](https://www.w3.org/TR/xml-names/#dt-NSName), the prefix should be an [NCName](https://www.w3.org/TR/xmlschema-2/#NCName) which allows dashes.

repro:
https://github.com/instant-labs/instant-xml/commit/11702ec4bbb6e0d1167217cb346de39ec2a04e70

djc:
Curious what you were up to!

Fixes #54.

### DIFF
--- a/instant-xml-macros/src/meta.rs
+++ b/instant-xml-macros/src/meta.rs
@@ -134,7 +134,9 @@ impl NamespaceMeta {
                 (NsState::Prefix, TokenTree::Ident(id)) => NsState::Eq {
                     prefix: id.to_string(),
                 },
-                (NsState::Eq { mut prefix }, TokenTree::Punct(punct)) if punct.as_char() == '-' => {
+                (NsState::Eq { mut prefix }, TokenTree::Punct(punct))
+                    if punct.as_char() == '-' || punct.as_char() == '.' =>
+                {
                     prefix.push(punct.as_char());
                     NsState::Eq { prefix }
                 }
--- a/instant-xml-macros/src/meta.rs
+++ b/instant-xml-macros/src/meta.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 use std::fmt;
 
-use proc_macro2::{Delimiter, Group, Ident, Literal, Punct, Span, TokenStream, TokenTree};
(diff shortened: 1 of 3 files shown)


instant-xml-macros/src/meta.rs
```diff
@@ -128,10 +128,20 @@ impl NamespaceMeta {
...
                     NsState::PrefixValue {
-                        prefix: segment.ident,
+                        prefix: segment.ident.to_string(),
                     }
                 }
-                (NsState::Prefix, TokenTree::Ident(id)) => NsState::Eq { prefix: id },
+                (NsState::Prefix, TokenTree::Ident(id)) => NsState::Eq {
+                    prefix: id.to_string(),
+                },
+                (NsState::Eq { mut prefix }, TokenTree::Punct(punct)) if punct.as_char() == '-' => {
```
nrempel, line 137: I think this should allow `_` and `.` as well.

I'm not sure if there is value in modelling the constraints as well but I believe the first character has to be ascii or `_` and the string cannot contain `xml` or `:`.

nrempel (commented):
Thanks! I took a shot at solving this last night but I haven't worked with the syn crate and I went down a rabbit hole!

nrempel:
I found a huge open dataset that I couldn't help be play with with 😝

https://www.sec.gov/edgar/browse/?CIK=320193&owner=exclude
https://www.sec.gov/Archives/edgar/data/320193/000032019324000006/aapl-20231230_htm.xml

instant-xml-macros/src/meta.rs
djc: Added support for `.` as well. The code is matching on an Identifier token in the token stream which will be a little more lenient than what you're describing, it should already fail on `:` (by virtue of the parser failing closed) and I think it's probably okay not to model the `xml` constraint to allow Rust types to model elements/attributes that are defined by XML standards.

nrempel (approved):

djc:
Published instant-xml-macros v0.4.2 at registry `crates-io`.

PR #58, opened 2024-04-06
Add support for comments

nrempel:
This PR adds support for comments.

https://www.w3.org/TR/xml/#sec-comments

### DIFF
--- a/instant-xml/src/de.rs
+++ b/instant-xml/src/de.rs
@@ -353,6 +353,7 @@ impl<'xml> Iterator for Context<'xml> {
                     false => return Some(Err(Error::UnexpectedToken(format!("{token:?}")))),
                     true => {}
                 },
+                Ok(Token::Comment { .. }) => continue,
                 Ok(token) => return Some(Err(Error::UnexpectedToken(format!("{token:?}")))),
                 Err(e) => return Some(Err(Error::Parse(e))),
             }
--- a/instant-xml/tests/direct.rs
+++ b/instant-xml/tests/direct.rs
@@ -2,7 +2,7 @@ use similar_asserts::assert_eq;
 
 use instant_xml::{from_str, to_string, FromXml, ToXml};
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Clone, Debug, Eq, FromXml, PartialEq, ToXml)]
 struct Foo {
     #[xml(attribute)]
     flag: bool,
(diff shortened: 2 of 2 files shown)


instant-xml/src/de.rs
```diff
@@ -108,17 +108,36 @@ impl<'xml> Iterator for Deserializer<'_, 'xml> {
...
+                        return Some(Err(Error::UnexpectedState("close element mismatch")));
+                    }
 
-        if self.context.stack.len() == self.level && local == self.local && prefix == self.prefix {
-            self.done = true;
-            return None;
+                    // If the closing element matches the current element being deserialized,
+                    // mark the deserialization as done and return None to indicate the end of iteration.
+                    self.done = true;
+                    return None;
```
djc, line 126: Would prefer to outline the `Close` case from the `match` at least, similar to how it was before.

instant-xml/src/de.rs
```diff
@@ -108,17 +108,36 @@ impl<'xml> Iterator for Deserializer<'_, 'xml> {
...
+                }
+                Node::Comment(_) => {
+                    // If the current node is a comment, skip it and continue to the next node.
+                    continue;
```
djc, line 130: For both of these other arms, maybe ditch the `;` from the values and move the comments to outside/before the arm instead of inside it, which should allow rustfmt to format this much more concisely. (Also I think the comments make more sense "on" the arm instead of "in" the arm.)

instant-xml/src/de.rs
```diff
@@ -353,6 +372,9 @@ impl<'xml> Iterator for Context<'xml> {
                     false => return Some(Err(Error::UnexpectedToken(format!("{token:?}")))),
                     true => {}
                 },
+                Ok(Token::Comment { text, .. }) => {
+                    self.records.push_back(Node::Comment(text.as_str().into()));
```
djc, line 376: Actually, not sure we have a reason to preserve these? Couldn't we just skip them for now, which would avoid the need for the `Deserializer` changes?

djc (approved):

instant-xml/src/de.rs
nrempel: Ah yeah, good point...

Unless we also allow serializing /// comments in structs into the xml. Then I would say we should keep this to maintain the 1:1 mapping.

Do you think it makes sense to add ser support for comments?

djc: Unless you have a use case for it in the near future, I think we shouldn't worry about it for a while. KISS!

instant-xml/src/de.rs
```diff
@@ -353,6 +353,9 @@ impl<'xml> Iterator for Context<'xml> {
                     false => return Some(Err(Error::UnexpectedToken(format!("{token:?}")))),
                     true => {}
                 },
+                Ok(Token::Comment { .. }) => {
+                    continue;
```
djc, line 357: Ditch the `;` to make this more concise.

djc (approved):

nrempel:
Well that's a lot simpler 😛

PR #62, opened 2024-06-02
Allow empty strings, instead of treating them as missing

wez:
Note that this PR is based on https://github.com/instant-labs/instant-xml/pull/61, but doesn't depend upon it; I didn't want to just push this straight into that PR, and I didn't want to create too many more branches to reconcile the version of this crate that is pulled into my downstream project!

### DIFF
--- a/instant-xml-macros/src/de.rs
+++ b/instant-xml-macros/src/de.rs
@@ -234,6 +234,7 @@ fn deserialize_struct(
     let mut declare_values = TokenStream::new();
     let mut return_val = TokenStream::new();
     let mut direct = TokenStream::new();
+    let mut after_loop = TokenStream::new();
 
     let mut borrowed = BTreeSet::new();
     for (index, field) in fields.named.iter().enumerate() {
@@ -263,6 +264,7 @@ fn deserialize_struct(
             field_meta,
             &input.ident,
             &container_meta,
+            &mut after_loop,
         );
 
         if let Err(err) = result {
@@ -361,6 +363,7 @@ fn deserialize_struct(
                         node => return Err(Error::UnexpectedNode(format!("{:?} in {}", node, #ident_str))),
                     }
                 }
+                #after_loop
 
                 *into = Some(Self { #return_val });
                 Ok(())
@@ -401,6 +404,7 @@ fn deserialize_inline_struct(
     let mut declare_values = TokenStream::new();
     let mut return_val = TokenStream::new();
     let mut direct = TokenStream::new();
+    let mut after_loop = TokenStream::new();
 
     let mut borrowed = BTreeSet::new();
     let mut matches = TokenStream::new();
@@ -433,6 +437,7 @@ fn deserialize_inline_struct(
             field_meta,
             &input.ident,
             &meta,
+            &mut after_loop,
         );
 
         let data = match result {
@@ -548,6 +553,7 @@ fn named_field<'a>(
     mut field_meta: FieldMeta,
     type_name: &Ident,
     container_meta: &ContainerMeta,
+    after_loop: &mut TokenStream,
 ) -> Result<FieldData<'a>, syn::Error> {
     let field_name = field.ident.as_ref().unwrap();
     let field_tag = field_meta.tag;
@@ -598,6 +604,15 @@ fn named_field<'a>(
         let mut #val_name = <#no_lifetime_type as FromXml>::Accumulator::default();
     ));
 
+    if field_meta.direct {
+        declare_values.extend(quote!(
+            // We will synthesize an empty text node when processing a direct
+            // element. Without this, we can miscategorize an empty tag as
+            // a missing tag
+            let mut seen_direct = false;
+        ));
+    }
+
     let deserialize_with = field_meta
         .deserialize_with
         .map(|with| {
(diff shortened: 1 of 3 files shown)


wez:
once both this and #61 are merged, I think it's reasonable to cut a new release of instant-xml; I've been sitting on both of these changes for the past week or so and haven't run into other weird cases around them in that time.

wez:
Hmm, hold on merging this: I tweaked the implementation and tests to make the PR look nicer, but running the tests in my downstream project now fail. I'm investigating!

wez:
OK, the problem case was with an empty nested direct field being treated as missing:

```rust
#[derive(Debug, PartialEq, FromXml)]
struct ArtUri {
    #[xml(direct)]
    uri: String,
}

#[derive(Debug, PartialEq, FromXml)]
struct Container {
    art: Option<ArtUri>,
}

#[test]
fn container_empty_string() {
    assert_eq!(
        from_str("<Container><ArtUri></ArtUri></Container>"),
        Ok(Container {
            art: Some(ArtUri {
                uri: "".to_string()
            })
        })
    );
}
```

 I've adjusted the codegen for that case; it now looks like this; I've annotated the new sections with `HERE`:

```rust
    fn deserialize<'cx>(
        into: &mut Self::Accumulator,
        field: &'static str,
        deserializer: &mut ::instant_xml::Deserializer<'cx, 'xml>,
    ) -> ::std::result::Result<(), ::instant_xml::Error> {
        use ::instant_xml::de::Node;
        use ::instant_xml::{Accumulate, Error, FromXml, Id, Kind};
        enum __Elements {
            __Ignore,
        }
        enum __Attributes {
            __Ignore,
        }
        let mut __value0 = <String as FromXml>::Accumulator::default();
        let mut seen_direct = false; // HERE: control flag to remember if we saw the text node
        loop {
            let node = match deserializer.next() {
                Some(result) => result?,
                None => break,
            };
            match node {
                Node::Attribute(attr) => {
                    let id = deserializer.attribute_id(&attr)?;
                    let field = __Attributes::__Ignore;
                    match field {
                        __Attributes::__Ignore => {}
                    }
                }
                Node::Open(data) => {
                    let id = deserializer.element_id(&data)?;
                    let element = __Elements::__Ignore;
                    match element {
                        __Elements::__Ignore => {
                            let mut nested = deserializer.nested(data);
                            nested.ignore()?;
                        }
                    }
                }
                Node::Text(text) => {
                    seen_direct = true; // HERE: set control flag
                    let mut nested = deserializer.for_node(Node::Text(text));
                    <String>::deserialize(&mut __value0, "ArtUri::uri", &mut nested)?;
                }
                node => {
                    return Err(
                        Error::UnexpectedNode({
                            let res = ::alloc::fmt::format(
                                format_args!("{0:?} in {1}", node, "ArtUri"),
                            );
                            res
                        }),
                    );
                }
            }
        }
        // HERE: this is the #after_loop code, same as the Node::Text code but with an
        // implied empty string value
        if !seen_direct {
            let mut nested = deserializer.for_node(Node::Text("".into()));
            <String>::deserialize(&mut __value0, "ArtUri::uri", &mut nested)?;
        }
        *into = Some(Self {
            uri: __value0.try_done("ArtUri::uri")?,
        });
        Ok(())
    }
```

instant-xml/src/impls.rs
```diff
@@ -270,9 +270,9 @@ impl<'xml> FromXml<'xml> for String {
         }
 
         match deserializer.take_str()? {
-            Some(value) => *into = Some(value.into_owned()),
-            None => return Ok(()),
-        }
+            Some(value) => into.replace(value.into_owned()),
+            None => into.replace(String::new()),
```
djc, line 274: You've fixed the `String` impl here -- do we need similar fixes for `&str` and `Cow<'_, str>`?

instant-xml-macros/src/de.rs
```diff
@@ -624,10 +642,21 @@ fn named_field<'a>(
         } else if field_meta.direct {
             direct.extend(quote!(
                 Node::Text(text) => {
+                    seen_direct = true;
```
djc, line 645: Does this result in different handling between self-closing tags and empty text nodes? Is that desirable?

instant-xml/src/impls.rs
wez: Yeah; I've added `Cow<'_, str>` in the latest push. There is no impl for `str`.

instant-xml-macros/src/de.rs
wez: The behavior is the same for both; I've added a couple of test cases to demonstrate that.
I think having them behave consistently with each other is reasonable.
I don't need to make a distinction in my downstream code.

I imagine that if you did care about the difference, you would want the direct value to be `Option<String>`:

```rust
#[derive(Debug, PartialEq, FromXml)]
struct OptionArtUri {
    #[xml(direct)]
    uri: Option<String>,
}
```

with the current state in this PR, we'll produce `Ok(OptionContainer { art: Some(OptionArtUri { uri: Some("") }) })` for that case.

I'm not sure if there's a reasonable way to make that come back as `uri: None` without making more significant changes to the traits and types, and/or having some understanding about whether the target type is `Option`.

Without this PR, this case produces a `MissingValue` error, so I don't think that things are worse off than they were.

djc: Agree that this behavior is desired.

djc:
This makes sense to me! I rebased and made some minor style changes in the `impls` module.

djc (approved):

djc:
* Published instant-xml v0.5.0 at registry `crates-io`
* [new tag]         0.5.0 -> 0.5.0

PR #64, opened 2024-06-07
Provide more context in DuplicateValue errors

djc:


valkum (approved):
Might be worth to also add an UI tests for the macros using trybuild? At least an issue marked as easy for anyone to pick up. E.g. Hacktober or similar events.

djc:
> Might be worth to also add an UI tests for the macros using trybuild? At least an issue marked as easy for anyone to pick up. E.g. Hacktober or similar events.

Fair, filed https://github.com/instant-labs/instant-xml/issues/65.

PR #66, opened 2024-07-24
Update README.md

hartshorne:


djc (approved):

djc:
Could also throw in a mention of instant-epp.

PR #70, opened 2024-10-22
fix: support all accumulators in enum forward deserialization

gibbz00:
Previously limited to Option<T> only.

The following does not compile in main:

```rust
#[derive(Debug, PartialEq, ToXml, FromXml)]
#[xml(forward)]
enum FooCow<'a> {
    Bar(Cow<'a, [BarBorrowed<'a>]>),
    Baz(Cow<'a, [BazBorrowed<'a>]>),
}
```

### DIFF
--- a/instant-xml/tests/forward-enum.rs
+++ b/instant-xml/tests/forward-enum.rs
@@ -4,19 +4,19 @@ use similar_asserts::assert_eq;
 
 use instant_xml::{from_str, to_string, FromXml, ToXml};
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Debug, FromXml, PartialEq, ToXml)]
 #[xml(forward)]
 enum Foo {
     Bar(Bar),
     Baz(Baz),
 }
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Debug, FromXml, PartialEq, ToXml)]
 struct Bar {
     bar: u8,
 }
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Debug, FromXml, PartialEq, ToXml)]
 struct Baz {
     baz: String,
 }
--- a/instant-xml/tests/forward-enum.rs
+++ b/instant-xml/tests/forward-enum.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use similar_asserts::assert_eq;
 
 use instant_xml::{from_str, to_string, FromXml, ToXml};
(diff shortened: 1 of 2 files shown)


instant-xml/tests/forward-enum.rs
```diff
@@ -9,20 +11,50 @@ enum Foo {
...
+#[xml(rename = "Bar")]
+struct BarBorrowed<'a> {
+    bar: Cow<'a, str>,
+}
+
+#[derive(Clone, Debug, Eq, FromXml, PartialEq, ToXml)]
+#[xml(rename = "Baz")]
+struct BazBorrowed<'a> {
+    baz: Cow<'a, str>,
+}
```
djc, line 41: Nit: let's move these below the `wrapped_enum()` test.

Why do we need all the `Clone` derives? Do we actually need them?

djc (approved):
Nice work figuring this out, this crate isn't exactly well-documented...

Can you explain at a conceptual level why you want/need this?

instant-xml/tests/forward-enum.rs
gibbz00: > Nit: let's move these below the wrapped_enum() test.

Sure, fixed :) Any reason why you prefer to reduce the number of #[test]? Normally more inclined to do the opposite.

> Why do we need all the Clone derives? Do we actually need them?

Yes, I believe it's because of this.

https://github.com/djc/instant-xml/blob/0f641fdc1afaa4bd519741a31910b5e624e4e5c8/instant-xml/src/impls.rs#L330-L336

Other derives weren't needed, made sure to remove them.

instant-xml/tests/forward-enum.rs
```diff
@@ -4,19 +4,19 @@ use similar_asserts::assert_eq;
...
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Debug, PartialEq, FromXml, ToXml)]
 #[xml(forward)]
 enum Foo {
     Bar(Bar),
     Baz(Baz),
 }
 
-#[derive(Clone, Debug, Eq, FromXml, PartialEq, ToXml)]
```
djc: Prefer not to add stuff in one commit only to remove it again in a latter one.

(Though thanks for cleaning up the derives, that's a nice touch!)

instant-xml/tests/forward-enum.rs
djc: > > Nit: let's move these below the wrapped_enum() test.
> 
> Sure, fixed :) Any reason why you prefer to reduce the number of #[test]? Normally more inclined to do the opposite.

Sorry, I didn't mean "into" but "below", want to keep the types close to the test they belong with, so <types for test 1> <test 1> <types for test 2> <test 2> rather than <types for test 1> <types for test 2> <test 1> <test 2>.

> 
> > Why do we need all the Clone derives? Do we actually need them?
> 
> Yes, I believe it's because of this.
> 
> https://github.com/djc/instant-xml/blob/0f641fdc1afaa4bd519741a31910b5e624e4e5c8/instant-xml/src/impls.rs#L330-L336
> 
> Other derives weren't needed, made sure to remove them.

Okay, thanks.

gibbz00: Ok gotchu. Would I be taking it too far if I place the corresponding struct definitions inside the corresponding tests?

gibbz00:
> Nice work figuring this out, this crate isn't exactly well-documented...

Thanks! Yeah, at this stage it will have to be a dedicated effort for it to happen. Not a big deal right now since those looking this crate in the first place will probably have a fairly good understanding of serde, which helps.

> Can you explain at a conceptual level why you want/need this?

Sure! My task at work is to add support for secDNS per RFC 5910. (https://github.com/djc/instant-epp/issues/17). The `<info>` part returns the same `DsOrKeyData` enum currently used by `<create>`. It could be reused with a slight semver bump if:

```rust
#[derive(Debug, ToXml)]
#[xml(forward)]
pub enum DsOrKeyData<'a> {
    DsData(&'a [DsDataType<'a>]),
    KeyData(&'a [KeyDataType<'a>]),
}

// is turned into this:

#[derive(Debug, ToXml, FromXml)]
#[xml(forward)]
pub enum DsOrKeyData<'a> {
    DsData(Cow<'a, [DsDataType<'a>]>),
    KeyData(Cow<'a, [KeyDataType<'a>]>),
}
```

instant-xml/tests/forward-enum.rs
gibbz00: No worries, fixed.

instant-xml/tests/forward-enum.rs
```diff
@@ -4,19 +4,19 @@ use similar_asserts::assert_eq;
 
 use instant_xml::{from_str, to_string, FromXml, ToXml};
 
-#[derive(Debug, Eq, FromXml, PartialEq, ToXml)]
+#[derive(Debug, PartialEq, FromXml, ToXml)]
```
djc, line 7: Sorry, one more tiny nit: prefer to keep these in alphabetic order, as before.

djc (approved):

instant-xml/tests/forward-enum.rs
gibbz00: It's okay, normally I do too, but grouped by those from std and then by each crate.

Fixed.

gibbz00: Shame that order may sometimes matter, (bad derive macro implementation imo), otherwise it could be taken care of by rustfmt.

PR #71, opened 2024-10-22
Implement to and from xml for Box<T>

gibbz00:
The result of some experiments I did the other day, that themselves didn't lead to much.
Might as well post the addition here in case someone else finds the change useful.

### DIFF
--- a/instant-xml/src/impls.rs
+++ b/instant-xml/src/impls.rs
@@ -498,6 +498,32 @@ impl<T: ToXml + ?Sized> ToXml for Box<T> {
     }
 }
 
+impl<'xml, T: FromXml<'xml>> FromXml<'xml> for Box<T> {
+    #[inline]
+    fn matches(id: Id<'_>, field: Option<Id<'_>>) -> bool {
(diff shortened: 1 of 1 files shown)


instant-xml/src/impls.rs
```diff
@@ -488,6 +488,42 @@ impl<T: ToXml> ToXml for Option<T> {
...
+        field: &'static str,
+        deserializer: &mut Deserializer<'cx, 'xml>,
+    ) -> Result<(), Error> {
+        if into.is_some() {
+            return Err(Error::DuplicateValue(field));
+        }
+
+        let mut value = T::Accumulator::default();
+        T::deserialize(&mut value, field, deserializer)?;
+        *into = Some(Box::new(value.try_done(field)?));
```
gibbz00, line 518: Question, can the try operator be replaced with expect?

djc (approved):
LGTM, thanks!

instant-xml/src/impls.rs
djc: What do you mean?

PR #77, opened 2024-10-23
Prepare changes to the macro crate.

gibbz00:
Feedback from https://github.com/djc/instant-epp/pull/56 resulted in the
suggestion of at least 3 new potential macro attributes, (from, into, remote).
Before tackling those, however, I'd like to do some refactors and improvements
to the macro crate. Specifically in terms of making it clearer of when a
attribute belongs to a field, a container; when it is for serialization,
deserialization; and lastly, when it is for named, unnamed, or unit fields of
structs and enums. (There are at least 2 x 2 x 3 = 12 places where attributes
might be valid, not taking into account cases when different attributes are
incompatible with each other.)

Validation is scattered in the codebase, errors are seldom error emitted, and
there's little user documentation to rely on. This PR is the first one in a
series whose goal is to tackle exactly those 3 points. This once splits up some
larger files and uses `use crate::*;` to improve locality of behavior.

To clarify, this PR changes no behavior whatsoever.


instant-xml-macros/src/lib.rs
```diff
@@ -27,228 +24,6 @@ pub fn from_xml(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     de::from_xml(&ast).into()
 }
 
-pub(crate) use container::{ContainerMeta, Mode};
```
djc: Not a fan of this change. IMO 100-200 line files are a net negative to my ability to understand the code, my optimal size is around 300-800 lines of non-test code.

instant-xml-macros/src/meta/namespace.rs
```diff
@@ -0,0 +1,374 @@
+pub use base::Namespace;
+mod base {
```
djc, line 2: I don't think these inline modules are an improvement. Extracting `namespace` out of `meta` makes sense to me, but leaves very little code in `meta` itself... maybe it then makes sense to move the remainder into `lib.rs`?

instant-xml-macros/src/de.rs
```diff
@@ -4,9 +4,7 @@ use proc_macro2::{Ident, Literal, Span, TokenStream};
 use quote::quote;
 use syn::spanned::Spanned;
 
-use crate::meta::{ContainerMeta, FieldMeta, Mode, VariantMeta};
-
-use super::{discard_lifetimes, meta_items, Namespace};
+use crate::*;
```
djc, line 7: I don't like * imports, except for the single case of inline `tests` modules using `super::*`.

instant-xml-macros/src/lib.rs
gibbz00: Understand. For me it a visibility modefier reason to limit types and their impl blocks to their own module, with quite a few exceptions of course. Easier to reason about them then. Save thing as with star imports, ok if i flatten things in the end?

instant-xml-macros/src/meta/namespace.rs
gibbz00: I would personally want to extract those parts into their own module for now, suspect they will disappear once I'm finished. Thus having a mod.rs files that declares modules and exports only.

instant-xml-macros/src/de.rs
djc: Sorry, no. If you want me to review stuff, please optimize for reviewing as I suggest.

instant-xml-macros/src/lib.rs
djc: Similarly to the other aspect, I don't want this right now. If you end up adding substantial logic for one of these types we can definitely consider splitting it out at that point.

instant-xml-macros/src/meta/mod.rs
```diff
@@ -0,0 +1,14 @@
+mod container;
```
djc, line 1: IMO this is still overindexing on small modules. I think it would make sense to move `meta.rs` to `meta/mod.rs`, and only split out `namespace` into a separate file for now.

PR #78, opened 2024-10-24
chore: add `rustfmt` import rules

gibbz00:
In response to: https://github.com/djc/instant-xml/pull/77#discussion_r1814758755

### DIFF
--- a/instant-xml-macros/src/case.rs
+++ b/instant-xml-macros/src/case.rs
@@ -5,7 +5,6 @@
 // See https://users.rust-lang.org/t/psa-dealing-with-warning-unused-import-std-ascii-asciiext-in-today-s-nightly/13726
 #[allow(deprecated, unused_imports)]
 use std::ascii::AsciiExt;
-
 use std::fmt::{self, Debug, Display};
 
 use proc_macro2::Ident;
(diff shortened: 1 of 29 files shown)


gibbz00:
I do prefer `imports_granularity = "Crate"`, module seemed to be used slightly more.

djc:
I don't think this is much of an improvement.

- I usually stable `cargo fmt`, and I think most people do, so the configuration will just be ineffective and add a bunch of warnings, and will not be checked in CI.
- While technically `instant_xml` is external in its own integration tests, that does not feel conceptually correct to me, making most of the changes in this PR easier to technically enforce but also conceptually kinda wrong.
- As such, I'd prefer a PR that just takes the other changes here without adding configuration, relying more on reviewing to catch occasional lapses.

Code is for people -- tooling can help but IMO it is unavoidable that some stylistic aspects will remain eeasier to enforce by humans. If you prefer I can add a `CONTRIBUTING.md` with extensive style policy (I would pretty much copy rustls' `CONTRIBUTING.md` which I mostly wrote).

djc:
Given the lack of follow-up, assuming you've forked this crate, so going to close this for now.

PR #87, opened 2025-01-30
Impl (de)serialization for chrono::NaiveDateTime

SirBrucey:
Adds implementations of `ToXml` and `FromXml` for `chrono::NaiveDateTime`.

### DIFF
--- a/instant-xml/tests/chrono.rs
+++ b/instant-xml/tests/chrono.rs
@@ -1,14 +1,16 @@
 #![cfg(feature = "chrono")]
 
-use chrono::{DateTime, TimeZone, Utc};
+use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use similar_asserts::assert_eq;
 
 use instant_xml::{from_str, to_string, FromXml, ToXml};
 
 #[derive(Debug, Eq, PartialEq, FromXml, ToXml)]
-struct Test {
-    dt: DateTime<Utc>,
+struct Test<T> {
+    dt: T,
 }
+
+type TestUtcDateTime = Test<DateTime<Utc>>;
 
 #[test]
 fn datetime() {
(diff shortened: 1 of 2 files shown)


instant-xml/tests/chrono.rs
```diff
@@ -1,23 +1,36 @@
...
 use instant_xml::{from_str, to_string, FromXml, ToXml};
 
 #[derive(Debug, Eq, PartialEq, FromXml, ToXml)]
-struct Test {
-    dt: DateTime<Utc>,
+struct Test<T> {
+    dt: T,
 }
 
+type TestUtcDateTime = Test<DateTime<Utc>>;
```
djc, line 13: Can you drop the test aliases? I don't think they help much here.

SirBrucey:
> Thanks! Would be interested to hear what you are using instant-xml for!

I am currently re-writing the EPP service we use at work in rust and came across your instant-epp crate (thanks you for your work!).  
One of the registrars (Nominet) we talk to sends all their datetimes in non UTC format, which meant I needed to add custom response types for NaiveDateTime's.  
I am happy to upstream this work once I have it stable, if you are interested?

We also talk to some broadband suppliers via XML directly, so in the future I can see us re-writing those services and using instant-xml directly.

instant-xml/tests/chrono.rs
SirBrucey: Sure

djc:
Thanks!

djc:
> One of the registrars (Nominet) we talk to sends all their datetimes in non UTC format, which meant I needed to add custom response types for NaiveDateTime's. I am happy to upstream this work once I have it stable, if you are interested?

Depends on how custom it is -- is this stuff allowed by the RFCs/schemas?

Issue #59, opened 2024-05-21
Correct way to emit namespace prefix on a container?


wez:
Thank you for building this crate!
I'm doing some stuff that unfortunately requires doing SOAP from some generated code.
I'm currently looking at replacing this bit of manual xml generation with `ToXml` enabled types:

```rust
       let body = format!(
            r#"
            <s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"
                s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
                <s:Body>
                    <u:{action} xmlns:u="{service}">
                        {payload}
                    </u:{action}>
                </s:Body>
            </s:Envelope>"#,
```

I'm having a difficult time figuring out the `ns` attribute functionality; I've been looking at the test cases to get a feel for it.

This is what I've come up with, which seems to emit the correct text, but I'm not sure that this is the correct way to do it:

```rust
#[derive(ToXml, Debug, Clone, PartialEq)]
#[xml(
    rename = "u:Stop",
    ns("", u = "urn:schemas-upnp-org:service:AVTransport:1")
)]
pub struct StopRequest {
    #[xml(rename = "InstanceID")]
    pub instance_id: u32,
}

const SOAP_ENCODING: &str = "http://schemas.xmlsoap.org/soap/encoding/";
const SOAP_ENVELOPE: &str = "http://schemas.xmlsoap.org/soap/envelope/";

#[derive(Debug, Eq, PartialEq, ToXml)]
#[xml(rename="s:Envelope", ns(
    "",
    s = SOAP_ENVELOPE
))]
struct Envelope<T: ToXml> {
    #[xml(attribute, rename = "s:encodingStyle")]
    encoding_style: &'static str,
    body: Body<T>,
}

#[derive(Debug, Eq, PartialEq, ToXml)]
#[xml(rename = "s:Body")]
struct Body<T: ToXml> {
    payload: T,
}

let action = Envelope {
    encoding_style: SOAP_ENCODING,
    body: Body {
        payload: StopRequest { instance_id: 0 },
    },
};

assert_eq!(
    instant_xml::to_string(&action).unwrap(),
    "<s:Envelope xmlns:s=\"http://schemas.xmlsoap.org/soap/envelope/\" \
        s:encodingStyle=\"http://schemas.xmlsoap.org/soap/encoding/\">\
        <s:Body>\
[...]

PR #89, opened 2025-02-27
Enable early pushing of namespaces

ch3pjw:
I need write a SOAP `Envelope` tag that is prefixed using a namespace prefix that it itself declares:
```xml
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">...</soap:Envelope>
```
This looks like the same thing as issue #59. I don't believe there's a way to get the `Serializer` to output the above as it currently stands.

This PR just defers writing out the namespace declaration if the `Serializer.state` shows we have not yet written the start tag. 
There might be a better way to do this!

NB: we make no effort to expose this in the derive macros!

I also note in an additional test that the `Serializer` state doesn't prevent some classes of errors, like trying to close tags in the wrong order. The same will be true of context `pop()`s happening in the wrong order, because the library doesn't use an internal stack in the `Serializer` but relies upon the client code passing stuff back in the correct order. Is that intentional for performance reasons?

### DIFF
--- a/instant-xml/src/ser.rs
+++ b/instant-xml/src/ser.rs
@@ -223,3 +223,28 @@ enum State {
     Element,
     Scalar,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use similar_asserts::assert_eq;
+
+    /// This test demonstrates that the state machine in the [`Serializer`] does not protect
+    /// library users from incorrectly nesting XML elements.
+    #[test]
+    fn detect_incorrectly_nested_tags() -> Result<(), Error> {
+        let mut s = String::new();
+        let mut ser = Serializer::new(&mut s);
+        let prefix_outer = ser.write_start("outer", "")?;
+        ser.end_start()?;
+        let prefix_inner = ser.write_start("inner", "")?;
+        ser.end_start()?;
+        ser.write_close(prefix_outer, "outer")?;
+        ser.write_close(prefix_inner, "inner")?;
+        // FIXME: This is a Bad Thing(TM) - we should really have a stack of tag names, instead of
+        // relying on the client code to pass us stuff in the correct order.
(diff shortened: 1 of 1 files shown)


djc:
Why do you need this?

ch3pjw:
The example I quoted is how the documentation for the API in interfacing with wrote out its example XML. I couldn’t see a way to do it with `instant-xml` without these changes.

djc:
I suppose my point is that you need to reproduce the XML data model, but that doesn't mean you need to reproduce the exact bytes that the example has. This crate is about bridging the abstract data model to the Rust type system, being able to produce arbitrary serialized XML is not a goal.

So, can you write an `Envelope` struct that gets accepted by your peer, or not?

ch3pjw:
Doesn't this example show that there's something in the XML data model that this library can't currently represent? I haven't looked at the deserialisation side yet - I wanted to test the water with this change first - but I presume it will equally not work with the example XML above.

I'm in the early stages of prototyping my API calls from docs, so I guess I can just deviate from their example and hope it all works. However, I can't control what comes back from the server, and their example documents are obviously structured as described.

djc:
If instant-xml isn't able to decode this correctly, I'll definitely consider it a bug! Also happy to hear if you're having problems encoding your SOAP payloads in a way that your peer systems accept.

ch3pjw:
It looks like we do have enough information in `FromXml::matches` to determine that we've got the right tag, so I guess I'll close for now and re-open if we have any issues sending the differently laid out XML to the server. (Plus we might experiment with a different XML library.)

djc:
I hope I didn't turn you off with my response to your PR -- I do approach the attempt to upstream changes. I just want to limit complexity by sticking with the underlying data model. Happy to help you address interoperability problems with your SOAP implementation!

PR #91, opened 2025-03-06
Improve handling of optional direct fields

djc:
@vshashi01 hope this solves your issues, do you want to try it with your real-world code?


vshashi01:
@djc Yes this fixes my issue on main project. Question, do I understand correctly the current fixes, only fixes for when the direct value is None value right? Would this change also be applicable for xml element that only has attribute fields without Element fields?

djc:
> Question, do I understand correctly the current fixes, only fixes for when the direct value is None value right? Would this change also be applicable for xml element that only has attribute fields without Element fields?

Write some more tests?

PR #92, opened 2025-03-07
Use empty element tags for attribute-only types

djc:
@vshashi01 does this cover what you meant in https://github.com/djc/instant-xml/pull/91#issuecomment-2705120614?


vshashi01:
I will check and get back to you later today.

vshashi01:
Was able to validate quickly, yes this is works!

djc:
Okay, let me know if you'd like a release or file new issues if you encounter more issues/have more questions!

PR #93, opened 2025-03-09
Failure on namespaced enum as attribute

vshashi01:
When we add a namespaced enum as an attribute, the deserializer is unable to detect the field at all. 

The unit test to show serializer works but the deserializer failed with MissingValue(StructWithNamespacedEnumAttr::scalar).

Adding the enum as standard field works correctly.

### DIFF
--- a/instant-xml/tests/de-enum-attr-ns.rs
+++ b/instant-xml/tests/de-enum-attr-ns.rs
@@ -1,0 +1,46 @@
+use similar_asserts::assert_eq;
+
+use instant_xml::{from_str, to_string, FromXml, ToXml};
+
+const NS: &str = "\\Some\\Namespace";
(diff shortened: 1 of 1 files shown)


djc:
I think this is working as intended? A `scalar` enum acts as a scalar, not an element (so, represented by a string). As such, it cannot have a namespace. In this case (any case where the value is a scalar), you can stick the namespace on the field.

vshashi01:
Yeap assigning to the field makes it possible for my use case. But I discounted the fact with scalar its anyway not possible to specify the namespace.

djc:
There is another enum mode IIRC which you maybe could use.

vshashi01:
I see that Scalar is the only mode eligible for enums right? Once I am done with my current project, will try to contribute on documentation.

djc:
Nope, there's also `forward` mode. Have a look at the `forward-enum` test.

PR #94, opened 2025-03-11
Make namespace paths consistently refer to full namespace

djc:
`FromXml` and `ToXml` treated `ns(foo)` on fields inconsistently:

- `FromXml` treated `foo` as a path to a constant containing the full namespace string
- `ToXml` treated `foo` as a namespace prefix defined on the container type

Change the `ToXml` treatment to match the `FromXml` treatment, favoring less indirection.

Use a semver-incompatible version bump to signal the potential incompatibility.

cc @vshashi01

### DIFF
--- a/instant-xml-macros/Cargo.toml
+++ b/instant-xml-macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-xml-macros"
-version = "0.5.2"
+version = "0.6.0"
 edition = "2021"
 rust-version = "1.61"
 workspace = ".."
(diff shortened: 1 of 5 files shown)


vshashi01:
Yeap this works!

vshashi01:
I think the deserialization still fails with a UnknownPrefix error. Extending the test to include a FromXml test should reveal this I belief.

djc:
> I think the deserialization still fails with a UnknownPrefix error. Extending the test to include a FromXml test should reveal this I belief.

Nice catch! Fixed it.

PR #98, opened 2025-07-22
Fix creator of serde

sophiajt:
Fixing the serde creator to the correct creator. While dtolnay maintains serde now, he didn't create it. It was created by erickt.

https://github.com/serde-rs/serde/graphs/contributors


djc (approved):
Thanks!

Issue #95, opened 2025-03-18
Get all the none default Namespaces and its Prefixes found in the XmlDocument on deserialization


vshashi01:
As a user, I want the ability to extract all the none default namespaces and prefixes encountered in the current Xml document that is deserialized.

Its very common for the producers of Xml document to encode additional data in some own namespace with its own custom prefix. As it stands, when using instant-xml API, the deserializer needs to know this namespace to be able to deserialize the relevant fields (which makes sense). 

However, sometimes we just want to be able to catch the namespaces unknown to the Deserializer itself so that we can use this on the application level for user feedback or other transformation of the deserialized data. 

``` rust

#[derive(ToXml, FromXml, PartialEq, Debug, Eq)]
#[xml(ns("Default", p = "SomeUnknown"))]
    pub struct SomeData {
    pub field: String,
    
    #[xml(ns("SomeUnknown"))]
    pub newfield: String
}

#[test]
pub fn fromxml_test() {
    let xml_string = "<SomeData xmlns=\"Default\" xmlns:p=\"SomeUnknown\" xmlns:c=\"AnotherUnknown\"><field>Trial String field</field><p:newfield>NewField</p:newfield><c:anotherfield>AnotherField</c:anotherfield></SomeData>";
    let original_data = SomeData {field: "Trial String field".to_owned(), newfield: "NewField".to_owned()};
    let mut ns_map= HashMap::<String, String>::new();
    let data = from_str::<SomeData>(xml_string, &mut ns_map).unwrap();

    assert!(ns_map.len() == 2);
    assert_eq!(data, original_data);
}
 
```

I don't currently see a way to get this information, but since internally the Context stack is maintained up the chain, I imagine a way to extract this either through a new public function of from_str kind that can also allow us to pass in a mutable container to get this info could be useful. 

But I am also open to other suggestion for a cleaner API as well.

djc:
I'd try something like this:

```rust
struct UnknownElements(HashMap<(String, String), String>);

impl<'xml> FromXml<'xml> for UnknownElements {
     fn matches(id: Id<'_>, field: Option<Id<'_>>) -> bool {
          true // or maybe false only for `id.ns` values that you know about?
     }

     fn deserialize<'cx>(
         into: &mut Self::Accumulator,
         field: &'static str,
         deserializer: &mut Deserializer<'cx, 'xml>,
     ) -> Result<(), Error> {
          // Left as an exercise to you... 
     }
}
```

[...]

PR #105, opened 2025-09-15
Add the ability to get all the namespaces found in the de-serialized XML document

vshashi01:
When looking at the proposed solution for issue #95, I realized it does not work for me because what I want to achieve is to get a full list of namespaces and the corresponding prefixes defined for the de-serialized XML doc. 

Based on that, the most obvious solution is to be able to track and return the namespaces encountered by de-serializer's Context object as it de-serializing the document.  

## Public API 
- Added a new de-serializer function; from_str_with_namespace() that returns the de-serialized object and a map of all the namespaces (BTreeMap<String, String>) found when de-serializing the XML document. 

## Internals
- Added a new field to the de-serializer Context to track all the namespaces
- When the namespaces are encountered during parsing of attribute Tokens, the namespace and its prefix is appended to the map. 
- Note: currently the implementation is basic only such that the last encountered prefix for the same namespace is the tracked prefix. It does not keep track of multiple prefixes that point to the same namespace URI.  
- Added a crate level pub function for de-serializer Context, to get a owned map of all of the namespaces from lib.rs

## Test
- Added tests to test that the new function returns the correct namespaces
- There are still couple of tests failing that I would need to investigate further and probably push some fixes.

## Potential improvements to the current approach 
- Unify the from_str and from_str_with_namespaces function to use the same underlying function so that the de-serializing correctness does not need to be retested per function (Note: Currently the added tests do not test for the correctness of de-serialization itself, only the namespaces, ideally if we keep two unique implementation for deserializing we should extend the tests to both.)
- If the new API adds unnecessary noise to default public API, it should also be possible to isolate the function and all the associated codes to a separate cargo feature such as "namespaces-tracking"

I am also open to an alternative suggestion on how we can achieve similar behavior with a different approach.

### DIFF
--- a/instant-xml/src/de.rs
+++ b/instant-xml/src/de.rs
@@ -125,6 +125,7 @@ pub(crate) struct Context<'xml> {
     parser: Tokenizer<'xml>,
     stack: Vec<Level<'xml>>,
     records: VecDeque<Node<'xml>>,
+    all_namespaces: BTreeMap<String, String>,
 }
 
 impl<'xml> Context<'xml> {
@@ -133,6 +134,7 @@ impl<'xml> Context<'xml> {
             parser: Tokenizer::from(input),
             stack: Vec::new(),
             records: VecDeque::new(),
+            all_namespaces: BTreeMap::new(),
         };
 
         let root = match new.next() {
@@ -158,6 +160,10 @@ impl<'xml> Context<'xml> {
             },
             name: element.local,
         })
+    }
+
+    pub(crate) fn namespaces(&self) -> BTreeMap<String, String> {
+        self.all_namespaces.clone()
     }
 
     fn attribute_id(&self, attr: &Attribute<'xml>) -> Result<Id<'xml>, Error> {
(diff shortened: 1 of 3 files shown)


vshashi01:
Thanks for looking into this. Indeed it does have extra overhed to be included by default, and currently its just an experiment to show it is possible to have this tracking. I am still confident its possible to make this a additive cargo feature that anyone who does not need it would be able to not incur the cost at all. I will work on refining this proposal. 

Meanwhile to give you more context on why this feature is important for me, 

The 3mf file specification relies on XML namespaces to define new features as separate additive specification on the core specification. the "required" and "recommended" extensions are expected to be encoded by producer of the file and be respected by the consumer of the file. 

The catch though, this field is usually populated with the prefix instead of the actual namespaces. 
image

 Example in a standard 3mf file:  
image

As i am writing a library for 3mf file reading, as a Consumer of such files, I should at the minimum be able to notify the users of the libraries of required extensions that are not currently supported by the library at runtime. But to deduce the required extension I would have to cross check the specified prefix to the actual namespace.

Hope this gives some context on why this feature would be useful for me.

vshashi01:
Thats clear but how do you see the Public API being? Would it still be by exposing this as 2 separate functions from_str and from_str_with_namespaces for example, or do you expect alternative Deserializer implementation to be provided by the consumer as some kinda generic parameter?

vshashi01:
I am going to close this pull request since I no longer think this is a necessary feature for this library at this time. I managed to localized the need for namespace tracking in my workflow and implement a minimal solution that works currently.

djc:
Sorry I wasn't able to help out with this.

vshashi01:
@djc No worries at all I also now think its unnecessary complexity for now. In the meantime, would it be possible to find some time to publish the current main to the crates.io? In crates.io, v0.5.2 is still the latest one.

djc:
I've published [0.6.0](https://github.com/djc/instant-xml/releases/tag/0.6.0).

PR #116, opened 2026-02-23
Add ValueElement for dynamic XML element capture

valkum:
This comes from https://github.com/djc/instant-epp/pull/88#issuecomment-3944690645 which requires support for 
```xml
<any namespace="##any" processContents="skip" />
```

This introduces a ValueElement type that captures arbitrary XML element trees at runtime, preserving namespace, name, attributes, text content, and nested children. This is useful for XML schemas that allow wildcard elements (xs:any) or when the element structure is not known at compile time.

This is the reading part of https://github.com/djc/instant-xml/issues/115. It does not cover serializing.

### DIFF
--- a/instant-xml-macros/Cargo.toml
+++ b/instant-xml-macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-xml-macros"
-version = "0.6.1"
+version = "0.6.2"
 edition = "2021"
 rust-version = "1.68"
 workspace = ".."
--- a/instant-xml/Cargo.toml
+++ b/instant-xml/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "instant-xml"
-version = "0.6.1"
+version = "0.6.2"
 edition = "2021"
 rust-version = "1.68"
 workspace = ".."


instant-xml/src/value_element.rs
```diff
@@ -0,0 +1,153 @@
...
+//! assert_eq!(parsed.inner.ns, "http://example.com");
+//! assert_eq!(parsed.inner.text.as_deref(), Some("text"));
+//! ```
+//!
+//! # Matching behavior
+//!
+//! `ValueElement` matches any XML element regardless of namespace or name. When
+//! used as a field in a derived struct, it will capture whichever element appears
+//! in that position. For capturing multiple arbitrary children, use
+//! `Vec<ValueElement>`.
```
djc: I don't think having this stuff as module-level docs makes a lot of sense, I think all of it should be on the type instead.

instant-xml/src/any_element.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Returns the local name of this element.
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    /// Returns the text content of this element, if any.
+    pub fn text(&self) -> Option<&str> {
+        self.text.as_deref()
+    }
```
djc: Nit: given that the fields are `pub` anyway, these getters seem pointless? Please remove them.

djc:
Addressing the CI MSRV failure in

- https://github.com/djc/instant-xml/pull/117

instant-xml/src/any_element.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Converts this element into an owned version with `'static` lifetime.
+    ///
+    /// This recursively converts all borrowed strings into owned copies,
+    /// decoupling the result from the original XML input.
+    pub fn into_owned(self) -> AnyElement<'static> {
```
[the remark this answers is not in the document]
valkum: Don't you order public to private anymore?

PR #122, opened 2026-03-16
Alternative force prefix

vshashi01:
- Added a new attribute for structs called force_prefix 
- Specifying force_prefix will force the current item to always force to define the local name with prefix and not requalify its default namespace in the start element
- Skipping the new attribute preserves the current default behavior (all new namespaces in the subtree are requalified as the default namespace on the first element in the hierarchy of the tree)
- Adding the attribute on enums, fields or inline structs, will be a compile error.

### DIFF
--- a/instant-xml/src/lib.rs
+++ b/instant-xml/src/lib.rs
@@ -157,6 +157,9 @@
 //!   let msg = Message::Request(Request {});
 //!   assert_eq!(to_string(&msg).unwrap(), "<Request />");
 //!   ```
+//!   
+//! -**`force_prefix`** *(structs only)* - Always serialize a namespace prefix if one is set for this element's namespace.
+//! Does not affect deserialization.
 //!
 //! ## Field attributes
 //!
--- a/instant-xml/src/ser.rs
+++ b/instant-xml/src/ser.rs
@@ -46,17 +46,39 @@ impl<'xml, W: fmt::Write + ?Sized> Serializer<'xml, W> {
             return Err(Error::UnexpectedState("invalid state for element start"));
         }
 
-        let prefix = match (ns == self.default_ns, self.prefixes.get(ns)) {
-            (true, _) => {
+        let force_prefix = cx.as_ref().is_some_and(|cx| cx.force_prefix);
+
+        let prefix = match (ns == self.default_ns, self.prefixes.get(ns), force_prefix) {
+            // Ns != default ns
+            (false, Some(prefix), force_prefix) => {
+                self.output.write_fmt(format_args!("<{prefix}:{name}"))?;
+                if let Some(cx) = &cx {
+                    // Only write the ns as default if not force prefix
+                    if !force_prefix {
+                        self.output
+                            .write_fmt(format_args!(" xmlns=\"{}\"", cx.default_ns))?;
+                    }
+                }
+                Some(*prefix)
+            }
+            // No prefix case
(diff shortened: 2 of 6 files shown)


djc (commented):
I think this mostly looks reasonable. Please squash all your commits.

instant-xml/src/ser.rs
```diff
@@ -46,17 +46,39 @@ impl<'xml, W: fmt::Write + ?Sized> Serializer<'xml, W> {
             return Err(Error::UnexpectedState("invalid state for element start"));
         }
 
-        let prefix = match (ns == self.default_ns, self.prefixes.get(ns)) {
-            (true, _) => {
+        let force_prefix = cx.as_ref().is_some_and(|cx| cx.force_prefix);
+
+        let prefix = match (ns == self.default_ns, self.prefixes.get(ns), force_prefix) {
+            (true, None, false) => {
+                //no prefix case
```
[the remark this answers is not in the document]
vshashi01: Updated all comments on the separate arms. 

Indeed `force_prefix` value for this case is irrelevant. Updated to ignore it.

instant-xml/src/ser.rs
```diff
@@ -46,17 +46,42 @@ impl<'xml, W: fmt::Write + ?Sized> Serializer<'xml, W> {
...
-                self.output.write_fmt(format_args!("<{prefix}:{name}"))?;
+            // If ns == default ns and force-prefix is false, we ignore prefix
+            (true, Some(_), false) => {
+                self.output.write_fmt(format_args!("<{name}"))?;
+                // Still requalify the ns here because its not the previous default_ns
                 if let Some(cx) = &cx {
-                    self.output
-                        .write_fmt(format_args!(" xmlns=\"{}\"", cx.default_ns))?;
+                    // only write the ns as default if not force prefix
+                    if !force_prefix {
```
djc: We should omit this condition, since we already know `force_prefix` is `false`, right?

instant-xml/src/lib.rs
```diff
@@ -168,6 +157,9 @@
 //!   let msg = Message::Request(Request {});
 //!   assert_eq!(to_string(&msg).unwrap(), "<Request />");
 //!   ```
+//!   
+//! -**`force_prefix`** *(structs only)* - If set does not requalify non-default namespace in the element declaration
```
djc: Reword as "Always serialize a namespace prefix if one is set for this element's namespace. Does not affect deserialization."

djc (approved):

instant-xml/src/ser.rs
vshashi01: Yes correct, missed it somehow! Done

instant-xml/src/lib.rs
vshashi01: Done.

djc (approved):

vshashi01:
@djc Any idea what is the timeline to publish a new version to crates.io?

djc:
Send me a PR with a version bump for both crates?

vshashi01:
Do you think a version bump is required still? Both the crates are already specified to be version 0.7.0 whereas the last released version was v0.6.2

djc:
Ah, in that case probably not.

djc:
[Published](https://github.com/djc/instant-xml/releases/tag/0.7.0).

PR #130, opened 2026-06-05
Potential reduction in unnessary allocation

vshashi01:
I was investigating further performance improvement in my library when I noticed that there were a pretty significant range of heap allocation and deallocation happening in the Context::next function. 

### Overall view of my single from_str call
image

### Zoomed in view of a single leading region with total execution %:
image

In my specific workload this allocation and deallocation made up consistenly 15-20% of the execution time. 

I am not an expert at XML decode, however, If we could just avoid the default String allocation on the case that the value does not contain any decode-able bytes, it seems it can already significantly improve speed in some workloads. 

### Process after adding an early return in `fn decode`

image

I have a consistent 15-18% improvement in speed across my benchmark suite from small(2KB) XML up to a large(600MB) XML deserialization. 

If this is not the best way to skip this allocation, feel free to provide guidance if there are alternatives I could try.

### DIFF
--- a/instant-xml/src/de.rs
+++ b/instant-xml/src/de.rs
@@ -399,6 +399,10 @@ pub fn borrow_cow_slice_u8<'xml>(
 }
 
 fn decode(input: &str) -> Result<Cow<'_, str>, Error> {
+    if !input.as_bytes().contains(&b'&') {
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice!

PR #131, opened 2026-06-05
de: lazily allocate for string decoding

djc:
@vshashi01 how does this compare in your benchmarks to your previous implementation?

Depending on how common it is to have long strings, a hybrid approach (with an early return only for short `input`s) might be better still, I suppose.

### DIFF
--- a/instant-xml/src/de.rs
+++ b/instant-xml/src/de.rs
@@ -399,11 +399,7 @@ pub fn borrow_cow_slice_u8<'xml>(
 }
 
 fn decode(input: &str) -> Result<Cow<'_, str>, Error> {
-    if !input.as_bytes().contains(&b'&') {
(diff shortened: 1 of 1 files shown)


vshashi01:
This looks cleaner indeed, let me pull and run my benchmarks tomorrow and let you know

vshashi01:
This looks like a cleaner improvement indeed, between the original version and this version flamegraph only shows a 1% variance in execution time within the iterator. 

## Original (by me)

### Flamegraph
image

### Overall execution times for my benchmarks:
image

## New version 

### Flamegraph
image

### Overall execution times for my benchmarks
image

I think we can say it is identical within run to run variance expected in performance.

djc:
Thanks for benchmarking!

PR #138, opened 2026-09-13
Add support for AnyAttribute and AnyElement serde

vshashi01:
My goal: To add support for serialization of deserialization of dynamic type like AnyAttribute and AnyElement 

What I found: 

- AnyAttribute was anyways not supported on deserialization to be able to capture Attributes that were already not matched by other attribute fields. 
- AnyElement only supported deserialization path but not serialization path. 

My approach: 

Introduce a special serialization and deserialization API for FromXml and ToXml which can be used in the derive macros for any field that is not previously matched anyway for attributes. 

Added a ToXml implementation for AnyElement as well. 

Do you see any other way to add this support to this library?


vshashi01:
Also another problem I dont have a solution to is how to write no static string to instant_xml::ser::Context with none static string for ns and prefixes.

djc:
I don't see why we need separate attribute serialize and deserialize methods?

PR #139, opened 2026-09-23
fix(macros): support deserialize_with on attribute fields

valkum:
This adds support for 
```
#[derive(Debug, FromXml, PartialEq)]
struct Bar {
    #[xml(attribute, deserialize_with = "deserialize_flag")]
    flag: bool,
}
```

This was missing until now. Test coverage should be good.


djc (approved):

djc:
Thanks!

PR #140, opened 2026-09-23
feat: improve derive around Option<Cow<str>> and Vec<Cow<str>>

valkum:
This is stacked on #139 and adds support for `Option<Cow<str>>` and `Vec<Cow<str>>`.

Previously, you needed to use a custom deserializer for borrow deserialization of `Option<Cow<str>>` and `Vec<Cow<str>>`. The default generated code allocated because the borrowing helper relied on `is_cow` which only matched `Cow<str>` not `Option<Cow<str>>`.

Because `Cow<str>` is already handled, I thought it might be good to add these two cases as well. `Option` and `Vec` are both part of the `std` lib and thus it makes sense to add first hand support in `instant-xml`. This should not set a precedent to add this support for all kinds of containers that might hold a `Cow`.

I also increased the test coverage around `Option<String>` to keep behavior consistent for the `String` and `Cow<str>` and also around lifetimes for borrowed items. I can pull these out into their own PRs if you want.


instant-xml/src/de.rs
```diff
@@ -442,6 +442,20 @@ pub fn borrow_option_cow_str<'a, 'xml: 'a>(
     Ok(())
 }
 
+/// Deserialize a borrowed `Vec<Cow<str>>` value
+///
+/// Helper function for deserializing `Vec<Cow<str>>` with zero-copy borrowing from the input.
+/// Each call appends one value. An empty element appends `""`, like the non-borrowing
+/// implementation.
+pub fn borrow_vec_cow_str<'a, 'xml: 'a>(
```
djc: Nit: suggest squashing this into the next commit.

instant-xml/tests/option.rs
```diff
@@ -75,3 +75,22 @@ fn option_borrow_element() {
...
     assert_eq!(de.maybe, None);
 }
+
+#[derive(Debug, Eq, FromXml, PartialEq)]
+struct Qux {
+    maybe: Option<String>,
+}
+
+#[test]
+fn option_borrow_empty_element() {
```
djc: Is this adding substantial coverage?

djc (commented):
(Please rebase on current `main`.)

instant-xml/tests/option.rs
valkum: This is making sure that the two cases (owned and borrowed) do not drift regarding the parsing of `<maybe/>` -> `Option("")`.

instant-xml/src/de.rs
valkum: Done

